PR #4, opened 2018-05-20
Assignment #1

VolhaAnishchanka:


HelgaShiryaeva/assignment1/Anagram
```diff
@@ -0,0 +1,63 @@
+import java.util.*;
+
+public class Anagram {
+    /**
+     * Supports punctuation marks : ?!,.:
```
brzozan: The convention is that at the beginning of the method comment there should be a short summary of what the method does, followed by more detailed information. This is quite important, especially that this method does something non-trivial - it does not just check if two strings are anagrams.

Could you please add summary?

Please refer to [javadoc documentation](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format), which has some good examples.

HelgaShiryaeva/assignment1/Anagram
```diff
@@ -0,0 +1,63 @@
...
+     * @param first is the first string to be detected if it is an anagram of
+     * @param second string
+     * @return true if strings are anagrams
+     */
+    public boolean areAnagrams(String first, String second) {
+        if(first.length() != second.length()) {
+            return false;
+        }
+        String PUNCTUATION_MARKS = " ,?!.:";
```
brzozan: Good use of constant to improve readability of code and reduce code duplication!

Typically such constants are created as `private static final` fields in the class. Could you please convert it to such field?

HelgaShiryaeva/assignment1/Anagram
```diff
@@ -0,0 +1,63 @@
+import java.util.*;
```
brzozan: It is safer to include only specific classes that you need instead of using asterisk.

First, it lets you avoid name conflicts in existing classes (`java.util.Date` vs `java.sql.Date` etc).

Second, imagine that you work on a large project and your class depends on two packages: `package1`, from which you need only class `A`, and `package2`, from which you use classes `B` and `C`. If your code does `import package1.*;` and `import package2.*` and then someone adds class named `B` to the first package, your code will break, even though there is no good reason for it.

HelgaShiryaeva/assignment1/Anagram
```diff
@@ -0,0 +1,63 @@
...
+
+public class Anagram {
+    /**
+     * Supports punctuation marks : ?!,.:
+     * @param first is the first string to be detected if it is an anagram of
+     * @param second string
+     * @return true if strings are anagrams
+     */
+    public boolean areAnagrams(String first, String second) {
+        if(first.length() != second.length()) {
```
brzozan: Coding convention in Java is that there is a whitespace between keyword and parenthesis that follows it. It might seem like a minor thing, but it makes code look more consistent and thus a bit easier to read. Could you please add spaces where needed?

Slightly outdated [Oracle doc](http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141388.html) and [Google Java coding conventions](http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141388.html) for reference.

HelgaShiryaeva/assignment1/Anagram
```diff
@@ -0,0 +1,63 @@
+import java.util.*;
+
+public class Anagram {
+    /**
+     * Supports punctuation marks : ?!,.:
+     * @param first is the first string to be detected if it is an anagram of
+     * @param second string
+     * @return true if strings are anagrams
+     */
+    public boolean areAnagrams(String first, String second) {
```
brzozan: The name of the method does not make it immediately obvious what the logic behind anagrams here is. If I saw this method, I would assume that it would return true for `"abc def"` and `"abcde f"` which is not the case.

Would some other name provide more information?

HelgaShiryaeva/assignment1/Anagram
```diff
@@ -0,0 +1,63 @@
...
+        }
+        List<String> firstStringWords = new ArrayList<>();
+        while(stringTokenizerFirst.hasMoreTokens()) {
+            firstStringWords.add(stringTokenizerFirst.nextToken());
+        }
+        while(stringTokenizerSecond.hasMoreTokens()) {
+            String wordFromSecondString = stringTokenizerSecond.nextToken();
+            boolean isAnagram = false;
+            for(String wordFromFirstString : firstStringWords) {
+                if(areWordsAnagrams(wordFromFirstString, wordFromSecondString)) {
```
brzozan: Optional optimization: `areWordsAnagrams` creates a `HashMap` with character counts for both words each time it is called - for each word from second string map will be created up to N times, where N is the number of words in first string.
Similarly, map for each word of first string will be created M times, where M is number of words in second string. 

Would there be a way to avoid that?

HelgaShiryaeva/assignment1/Anagram
```diff
@@ -0,0 +1,63 @@
...
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean areWordsAnagrams(String first, String second) {
+        Map<Character, Integer> characterCountMap = new HashMap<>();
+        for(char c : first.toCharArray()) {
+            if(characterCountMap.containsKey(c)) {
```
brzozan: Optional optimization: Java 8 has method [`getOrDefault(key, defaultValue)`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#getOrDefault-java.lang.Object-V-), which could be used to reduce this `if` to a single line.

HelgaShiryaeva/assignment1/Anagram
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+        for(char c : second.toCharArray()) {
+            Integer count = characterCountMap.get(c);
+            if(count == null || count == 0) {
+                return false;
+            }
+            characterCountMap.put(c, characterCountMap.get(c) - 1);
+        }
+        for(Map.Entry<Character, Integer> entry : characterCountMap.entrySet()) {
```
brzozan: Optional: could be simplified a tiny bit by iterating over [values](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#values--) instead of entries.

HelgaShiryaeva/assignment1/ReversedListAccessUtil
```diff
@@ -0,0 +1,91 @@
...
+            }
+        }
+        size++;
+    }
+
+    public void add(T data) {
+        insert(data, size);
+    }
+
+    private Node getNode(int index) {
```
brzozan: `getNode` (and methods that rely on it) behaves in an inconsistent way. Imagine that you have a list of size 1.

- `getNode(0)` will return head (correct).
- `getNode(1)` will also return head (not correct - head is at index 0).

HelgaShiryaeva/assignment1/ReversedListAccessUtil
```diff
@@ -0,0 +1,91 @@
...
+        Node(T value, Node next) {
+            this.next = next;
+            this.value = value;
+        }
+
+        T getValue() {
+            return this.value;
+        }
+
+        public void setValue(T value) {
```
brzozan: `setValue` is not really needed.

HelgaShiryaeva/assignment1/ReversedListAccessUtil
```diff
@@ -0,0 +1,91 @@
+public class ReversedListAccessUtil {
+    public <T> T getReversed(int index, MyList<T> list) {
+      return list.get(list.size() - index);
+    }
+}
+class MyList<T> {
```
brzozan: It would be better to move this class to a separate file, since it is part of the interface of your package and should be public - other packages cannot use `ReversedListAccessUtil.getReversed` if `MyList` is hidden as package private.

HelgaShiryaeva/assignment1/ReversedListAccessUtil
```diff
@@ -0,0 +1,91 @@
+public class ReversedListAccessUtil {
+    public <T> T getReversed(int index, MyList<T> list) {
```
brzozan: Name of the method does not really say what it does. Would some other name be better?

Also please add comment to provide more details.

HelgaShiryaeva/assignment1/ReversedListAccessUtil
```diff
@@ -0,0 +1,91 @@
+public class ReversedListAccessUtil {
```
brzozan: From the design perspective it would be better if this was not a separate class but instead `MyList` provided a method to access its elements from the end. What do you think?

HelgaShiryaeva/assignment1/ReversedListAccessUtil
```diff
@@ -0,0 +1,91 @@
...
+    public <T> T getReversed(int index, MyList<T> list) {
+      return list.get(list.size() - index);
```
brzozan: Clever use of size :). How would you implement it if you could not store size of the list?

brzozan (changes requested):
Thanks for the pull request! Good job!

Global comment: Java source files should be named Something.java - otherwise they will not compile. Could you rename them?

More detailed comments added to code.

HelgaShiryaeva/assignment1/Anagram.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.ArrayList;
+import java.util.StringTokenizer;
+
+public class Anagram {
+    private static final String PUNCTUATION_MARKS = " ,?!.:";
+    private Map<Character, Integer> characterCountMap;
```
brzozan: This field is being cleared and reused by areWordsAnagrams. In such cases it is better to have it as local variable rather than a field. Your code will work if this class is used by a single thread, but will break if two threads call areWordsAnagrams at the same time.

HelgaShiryaeva/assignment1/Anagram.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+        List<String> firstStringWords = new ArrayList<>();
+        while (stringTokenizerFirst.hasMoreTokens()) {
+            firstStringWords.add(stringTokenizerFirst.nextToken());
+        }
+        while (stringTokenizerSecond.hasMoreTokens()) {
+            String wordFromSecondString = stringTokenizerSecond.nextToken();
+            boolean isAnagram = false;
+            for (String wordFromFirstString : firstStringWords) {
+                if (areWordsAnagrams(wordFromFirstString, wordFromSecondString)) {
```
brzozan: Current code is not exactly wrong, since it is in line with what assignment asks for - sentences are anagrams if each word from one of them is an anagram of at least one word from the other sentence. The result of it is that sentences  "abc def" and "abc abc" will be considered anagrams.

What if we assumed stricter definition, that each word can be used only once in anagram check? i.e. "abc def" and "abc abc" would not be anagrams, but "abc def" and "fed cba" would.
What changes would need to be made?

HelgaShiryaeva/assignment1/Anagram.java
```diff
@@ -0,0 +1,78 @@
...
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean areWordsAnagrams(String first, String second) {
+        characterCountMap.clear();
+        for (char c : first.toCharArray()) {
+            Integer count = characterCountMap.getOrDefault(c, -1);
```
brzozan: This could be simplified to:
```
int count = characterCountMap.getOrDefault(c, 0);
characterCountMap.put(c, count + 1);
```

HelgaShiryaeva/assignment1/SignlyLinkedList.java
```diff
@@ -0,0 +1,80 @@
+package assignment1;
+
+
+public class SignlyLinkedList<T> {
+    private Node tail;
```
brzozan, line 5: This is a very creative solution to the problem :-) Not exactly what I had in mind, but it is absolutely not wrong.

brzozan (changes requested):
Thank you for making requested changes! This PR looks much better now. I added some more comments, but we are getting there.

Also: this is the first time I have seen such creative solution to the list problem :-)

HelgaShiryaeva/assignment1/Anagram.java
```diff
@@ -51,15 +51,11 @@ public boolean areAnagramsByWords(String first, String second) {
...
+    private boolean areWordsAnagrams(String first, String second, Map<Character, Integer> characterCountMap) {
         characterCountMap.clear();
         for (char c : first.toCharArray()) {
-            Integer count = characterCountMap.getOrDefault(c, -1);
-            if (count != -1) {
-                characterCountMap.put(c, count + 1);
-            } else {
-                characterCountMap.put(c, 1);
-            }
+            Integer count = characterCountMap.getOrDefault(c, 0);
```
brzozan: Just `int`. No need to keep the boxed type.

HelgaShiryaeva/assignment1/Anagram2.java
```diff
@@ -0,0 +1,87 @@
+package assignment1;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.ArrayList;
+import java.util.StringTokenizer;
+
+public class Anagram {
```
brzozan: No need to keep two files. Only one class with final version should be left in PR for merging.

HelgaShiryaeva/assignment1/Anagram.java
```diff
@@ -0,0 +1,74 @@
...
+    public boolean areAnagramsByWords(String first, String second) {
+        if (first.length() != second.length()) {
+            return false;
+        }
+        StringTokenizer stringTokenizerFirst = new StringTokenizer(first, PUNCTUATION_MARKS );
+        StringTokenizer stringTokenizerSecond = new StringTokenizer(second, PUNCTUATION_MARKS);
+        if (stringTokenizerFirst.countTokens() != stringTokenizerSecond.countTokens()) {
+            return false;
+        }
+        Map<Character, Integer> characterCountMap = new HashMap<>();
```
brzozan: `characterCountMap` is used only inside `areWordsAnagrams`. No need to create it outside that method and pass as parameter. Just create this map in the method where you use it.

HelgaShiryaeva/assignment1/Anagram.java
```diff
@@ -0,0 +1,74 @@
...
+                }
+            }
+            if (!isAnagram) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean areWordsAnagrams(String first, String second, Map<Character, Integer> characterCountMap) {
```
brzozan: This method does not really need `charactersCountMap` parameter and it would be good if it was public - someone using your class might be interested in simply checking if two words are anagrams without the whole sentence-related logic.

HelgaShiryaeva/assignment1/Anagram.java
```diff
@@ -0,0 +1,74 @@
...
+            characterCountMap.put(c, count + 1);
+        }
+        for (char c : second.toCharArray()) {
+            Integer count = characterCountMap.get(c);
+            if (count == null || count == 0) {
+                return false;
+            }
+            characterCountMap.put(c, characterCountMap.get(c) - 1);
+        }
+        for (Integer value : characterCountMap.values()) {
```
brzozan: Just `int`.

HelgaShiryaeva/assignment1/Anagram.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private boolean areWordsAnagrams(String first, String second, Map<Character, Integer> characterCountMap) {
+        characterCountMap.clear();
+        for (char c : first.toCharArray()) {
+            Integer count = characterCountMap.getOrDefault(c, 0);
+            characterCountMap.put(c, count + 1);
+        }
+        for (char c : second.toCharArray()) {
+            Integer count = characterCountMap.get(c);
```
brzozan: Just `int`. And if you use `getOrDefault(c, 0)` you do not need to check for `null`, since in both cases when you want to return `false` from the method value of `count` will be 0.

HelgaShiryaeva/assignment1/Anagram.java
```diff
@@ -0,0 +1,74 @@
...
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean areWordsAnagrams(String first, String second, Map<Character, Integer> characterCountMap) {
+        characterCountMap.clear();
+        for (char c : first.toCharArray()) {
+            Integer count = characterCountMap.getOrDefault(c, 0);
```
brzozan: Just `int`. Whenever you operate on simple things like numbers or booleans, 99% of the time you should use lower-case type (`int`, `long`, `boolean` etc).

brzozan (changes requested):
Thanks for making requested changes! The PR looks almost perfect :)

I added a few minor comments. After that your changes will be ready to merge.

HelgaShiryaeva/assignment1/Anagram.java
```diff
@@ -51,20 +50,20 @@ public boolean areAnagramsByWords(String first, String second) {
...
             characterCountMap.put(c, count + 1);
         }
         for (char c : second.toCharArray()) {
-            Integer count = characterCountMap.get(c);
-            if (count == null || count == 0) {
+            int count = characterCountMap.getOrDefault(c, 0);
+            if (count == 0) {
                 return false;
             }
             characterCountMap.put(c, characterCountMap.get(c) - 1);
```
brzozan: You already have count available in variable. Just `characterCountMap.put(c, count - 1)`.

brzozan (approved):
I added one small comment, but other than that this PR looks good to me. Please fix the comment and feel free to merge it. Thank you!

VolhaAnishchanka:
Yeeey, we've done it ! Thank you so much for all your comments. I start  to understand why code reviews are important 👍

PR #5, opened 2018-05-21
Assignment 1

YakimS:


YakimS/assignment1/Q1
```diff
@@ -0,0 +1,71 @@
+public class Assignment1Q1
```
brzozan, line 1: It would be better to give this class a name that would indicate what it does.

YakimS/assignment1/Q1
```diff
@@ -0,0 +1,71 @@
...
+
+    public static void main(String[] args)
+    {
+        String s1 = "si Apepl Apepl ppppo";
+        String s2 = "ppeAl ppopp Apepl is";
+        System.out.println(isSentenceAnagram(s1, s2,true));
+        System.out.println(s1 + "      " + s2);
+    }
+
+    static boolean isSentenceAnagram(String sentence1, String sentence2, boolean isCaseSensitive)
```
brzozan, line 12: Good name of the method that makes it immediately clear what to expect! It would be nice to add comment that provides a bit more details though - what does it mean that two sentences are anagrams?

To make this method usable for code outside of your package it should be made public. Could you make sure that methods intended for external use are public?

YakimS/assignment1/Q1
```diff
@@ -0,0 +1,71 @@
...
+    {
+        String[] sent1 = sentence1.split(" ");
```
brzozan, line 14: `sent1` and `sent2` do not really say much about what those variables are. What kind of data do they contain? What names would make it more obvious?

YakimS/assignment1/Q1
```diff
@@ -0,0 +1,71 @@
...
+
+        if(sent1.length != sent2.length) return false;
+
+        boolean[] isSent2WordAnagramFound = new boolean[sent2.length];
+
+        for(int i = 0 ; i < sent1.length ; i++)
+        {
+            for(int j = 0 ; j < sent2.length ; j++)
+            {
+                if (isSent2WordAnagramFound[j]) continue;
```
brzozan, line 25: Clever! It took me a while to understand what is the purpose of this line and that it is correct.

It would be good if you added some comments that would make it clear what is the algorithm supposed to do.

YakimS/assignment1/Q1
```diff
@@ -0,0 +1,71 @@
+public class Assignment1Q1
+{
+
+    public static void main(String[] args)
```
brzozan, line 4: Ideally, to make the class fully reusable and clean (by clean I mean not containing unnecessary code) this class should be split into two - one that contains main logic and separate one used to test it, in this case containing `main` method.

YakimS/assignment1/Q1
```diff
@@ -0,0 +1,71 @@
...
+    {
+        if(s1.length() != s2.length() ) return false;
+
+        if(!isCaseSensitive)
+        {
+            s1 = s1.toLowerCase();
+            s2 = s2.toLowerCase();
+        }
+
+        Hashtable<Character, Integer> hashtable = new Hashtable<>();
```
brzozan, line 53: Again, name of the variable does not really say what is it used for. Could you change it to something that will be more informative?

Also it is a bit better to user `HashMap` instead of `Hashtable` - it is a bit faster and more modern.

YakimS/assignment1/Q2
```diff
@@ -0,0 +1,89 @@
+public class Assignment1Q2 {
```
brzozan, line 1: Like in Q1, could you separate "business" part of code (`MyLinkedList` class) and the testing one (`main` method)?

YakimS/assignment1/Q2
```diff
@@ -0,0 +1,89 @@
...
+
+    public static void main(String[] args) {
+        MyLinkedList<Integer> myLinkedList = new MyLinkedList<>();
+        Integer[] ob = new Integer[10];
```
brzozan, line 5: This variable is not really needed. You can just do `myLinkedList.add(i)`.

YakimS/assignment1/Q2
```diff
@@ -0,0 +1,89 @@
...
+                currentNode = currentNode.getNext();
+            }
+
+            return (T) kthBackNode.getData();
+
+        }
+
+    }
+
+    public static class MyNode<T>
```
brzozan, line 63: `MyNode` is not used (and should not be used) anywhere outside `MyLinkedList`, so it would be better to make it a private inner class of `MyLinkedList`. Could you change code to do that?

YakimS/assignment1/Q2
```diff
@@ -0,0 +1,89 @@
...
+            ob[i] = i;
+            myLinkedList.add(ob[i]);
+            System.out.print(ob[i] + "");
+        }
+        System.out.println("");
+        System.out.print("kth ele:"+myLinkedList.getKthLastElement(9));
+    }
+
+
+    public static class MyLinkedList<T>
```
brzozan, line 17: This class as well as `MyNode` has really nice API. Good job!

Could you add some comments to explain what methods do?

YakimS/assignment1/Q2
```diff
@@ -0,0 +1,89 @@
...
+            }
+
+            MyNode kthBackNode = getHead();
+            while (currentNode.getNext() != null)
+            {
+                kthBackNode = kthBackNode.getNext();
+                currentNode = currentNode.getNext();
+            }
+
+            return (T) kthBackNode.getData();
```
brzozan, line 57: No need to cast. `getData()` already returns type `T`.

YakimS/assignment1/Q2
```diff
@@ -0,0 +1,89 @@
...
+
+        }
+
+    }
+
+    public static class MyNode<T>
+    {
+        private T data;
```
brzozan, line 65: `data` does not change within lifetime of `MyNode` object (there is no `setData` method), so it is a good practice to make such field `final`.

brzozan (changes requested):
Really nice PR! Good job!

Algorithms for both questions seem correct, but I have a few comments about structure and format.

Global comments:
- pull requests should contain valid Java code, i.e. files with .java extension, each containing at most one public class.
- it should be compilable - all needed library classes should be imported.
- code format should adhere to Google coding style (link is available in the code review presentation Laura shared with everyone after yesterday's training).

I also added a few more detailed comments in code.

YakimS/assignment1/Q1
YakimS: Sure. Is "AnagramValidityChecker" would do?

YakimS/assignment1/Q1
YakimS: No problem. All my next assignments will be documented, and I'll make sure to use right access modifier.

YakimS/assignment1/Q1
YakimS: Maybe "wordsOfSentence1", "wordsOfSentence2"?

YakimS/assignment1/Q1
YakimS: Thank you!
As promised, next time my code will be fully documented.

YakimS/assignment1/Q1
YakimS: Noted!

YakimS/assignment1/Q1
YakimS: Maybe "charsCountInWord"?
Good to know, off I go to learn more about the difference.

YakimS/assignment1/Q2
YakimS: Right, will do.

YakimS/assignment1/Q2
YakimS: True, Thanks!!

YakimS/assignment1/Q2
YakimS: Sure can and should.

YakimS/assignment1/Q2
YakimS: No problem

YakimS/assignment1/Q2
YakimS: Right. I'll delete it.

YakimS/assignment1/Q2
YakimS: Right! I forgot to add "final" because there was a setData() function which I deleted. Hope I'll notice it next time.

YakimS:
Thank you very much for the clear and helpful review!
I'm writing for myself all your offered corrections so I can make sure they will be implemented in the next assignment, including the request for .java file etc.
I wasn't invited to a training two days ago, but have one tomorrow, so ill make sure to get the presentation and the link so i can follow Google coding style.
Should i recommit the changes for further review?

brzozan:
Yes, please make changes to the files on current branch (just edit files in this PR; no need to create new branch). When you commit them, I will see it and will review code again. Thank you!

AnagramValidityChecker.java
```diff
@@ -0,0 +1,83 @@
...
+
+        for(int i = 0 ; i < wordsOfSentence1.length ; i++)
+        {
+            for(int j = 0 ; j < wordsOfSentence2.length ; j++)
+            {
+                // if current word has already found in the other sentence than
+                // we should'nt check it again.
+                if (isSent2WordAnagramFound[j]) continue;
+
+                // checks if the current word is anagram of the other, and mark it in isSent2WordAnagramFound in case one is the anagram of the other.
```
brzozan: Please break lines at some reasonable length. In this case new line before `isSent2WordAnagramFound` would be perfect. Shorter lines are easier to read.

AnagramValidityChecker.java
```diff
@@ -0,0 +1,83 @@
+package CodeUAss1Package;
+
+import java.util.HashMap;
+
+public class AnagramValidityChecker
+{
+    /* Given two strings, determine if one is an anagram of the other.
```
brzozan: Please follow javadoc conventions in class and method comments. Typical format is:
```
/**
 * Sentence with summary of what method does.
 *
 * <p>More sentences that clarify logic.
 */

File: AnagramValidityChecker.java [hunk 15]
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+
+        // store the amount of occurrence of each letter in the first word
+        HashMap<Character, Integer> charsCountInWord = new HashMap<>();
+        for(int i = 0 ; i < s1.length() ; i++)
+        {
+            char c = s1.charAt(i);
+
+            if(charsCountInWord.containsKey(c))    charsCountInWord.put(c, charsCountInWord.get(c)+1);
```
brzozan: This form of if is very difficult to read. Please follow Google coding style.

In general constructs like `if`, `for` etc should always be followed by block wrapped in `{...}`. This makes code much easier to read and less likely to introduce bugs when code is changed later.

brzozan (changes requested):
Thanks for making changes to the code, but please follow Google coding style.

AnagramValidityChecker.java
```diff
@@ -0,0 +1,83 @@
...
+    /* Given two strings, determine if one is an anagram of the other.
+   Two sentences are anagrams if each word in the resulting
+    sentence is an anagram of one of the words in the original sentence.*/
+    public static boolean isSentenceAnagram(String sentence1, String sentence2, boolean isCaseSensitive)
+    {
+        // split the string to the words they are containing
+        String[] wordsOfSentence1 = sentence1.split(" ");
+        String[] wordsOfSentence2 = sentence2.split(" ");
+
+        if(wordsOfSentence1.length != wordsOfSentence2.length) return false;
```
brzozan: It is much better to always stick to the same style of formatting, no matter if the block following `if` (or `for` etc) is one or more than one line.

With formatting like:
```
if (condition) {
  action();
}
```
it is immediately clear what code will be executed if condition is true.

Without braces and with code in a single line the pattern is broken and suddenly person who reads code needs to spend brain cycles on actively parsing code, for example to detect possible bugs like:
```
if (condition) action1(); action2();
```

[Serious bugs](https://www.imperialviolet.org/2014/02/22/applebug.html) were introduced in production code and lack of braces contributed to missing them in review.

KthElementLinkedListTester.java
```diff
@@ -0,0 +1,16 @@
+package CodeUAss1Package;
+
+
+public class KthElementLinkedListTester {
+
+    public static void main(String[] args) {
+        KthElementLinkedList<Integer> myLinkedList = new KthElementLinkedList<>();;
```
brzozan, line 7: Unnecessary `;` at the end.

brzozan (approved):
In general this PR looks good, so I will approve it, but please, please avoid single-line blocks after `if`/`for` without curly braces.
It is more difficult to read than with braces and is actually increases probability of missing bugs in review or introducing them during future code modifications.

YakimS:
Thank you!
I've documented all the main issues that you've reviewed and will make sure to take care of them in the next assignments, and in general.

PR #6, opened 2018-05-21
Narciuch Assigment 1

narciuch:


narciuch/assignment1/src/Anagram.java
```diff
@@ -0,0 +1,123 @@
+import java.util.Arrays;
+
+public class Anagram {
+
+  /**
+   * Used in isAnagram() and isAnagramSentence()
+   * to ignore all characters not in a-z / A-Z / space.
+   */
+  private static final String replaceRegex = "[^a-zA-z ]";
```
brzozan: Good use of `static final` field!

Minor comments:
- As per Google coding style constants should be named using capital letters, i.e. `REPLACE_REGEX`.
- Comment clarifies purpose of this constant, but it would still be nice to give it a name that makes it obvious without looking at comment. Could you rename it, please?

narciuch/assignment1/src/Anagram.java
```diff
@@ -0,0 +1,123 @@
...
+   * @return true if sentence2 is a sentence anagram of sentence1.
+   */
+  public boolean isAnagramSentence(String sentence1, String sentence2) {
+
+    String[] words1 = preProcessSentence(sentence1);
+    String[] words2 = preProcessSentence(sentence2);
+
+    //Compare word by word if each word in sentence2 is an anagram of the word in sentence1
+    for (int i = 0; i < words1.length; i++) {
+      if (!isAnagram(words1[i], words2[i])) {
```
brzozan, line 93: This logic assumes that sentences are anagrams only if corresponding words are anagrams, i.e. "abc def" and "cba fed" are anagrams, but "abc def" and "def abc" are not.

Original text in the assignment says "... where each word in the resulting sentence is an anagram of *one of the words* in the original sentence".

How could the code be changed to handle original requirement?

narciuch/assignment1/src/Anagram.java
```diff
@@ -0,0 +1,123 @@
...
+    } else {
+      return isAnagram(string1, string2);
+    }
+  }
+
+  /**
+   * Determines if one sentence is an anagram of another,
+   * i.e. each word in the resulting sentence is an anagram of one of the words
+   * in the original sentence.
+   * The order of letters within each word can be shuffled as well as the order of the words.
```
brzozan: Minor issue: typically an empty comment line and a start of paragraph (`<p>`) is inserted between paragraphs of javadoc, including first sentence. In this case it would be something like:

```
/**
 * Determines if ...
 *
 * <p>The order...
 *
 * <p>Input strings are unchanged...
 */
```

narciuch/assignment1/src/Anagram.java
```diff
@@ -0,0 +1,123 @@
...
+    String[] words = string.split("\\s+");
+
+    //Sort characters within each word
+    for (int i = 0; i < words.length; i++) {
+      char[] chars = sortCharsInString(words[i]);
+      words[i] = new String(chars);
+    }
+
+    //Sort the entire word array
+    Arrays.sort(words);
```
brzozan: Clever optimization! It is not strictly required for correctness, so it might be worth it to mention why this sorting is done in the comment.

narciuch/assignment1/src/SinglyLinkedList.java
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Finds the kth to last element of the list
+   *
+   * @param nrFromEnd: desired index from end of list.
+   *                   nrFromEnd of 0 corresponds to the last element in the list,
+   *                   nrFromEnd of 1 corresponds to the second to last element etc.
+   * @return value of the nrFromEnd element
+   */
+  public T getKthElemFromTail(int nrFromEnd) {
+    return getElemAtIndex(size - nrFromEnd - 1);
```
brzozan, line 36: Clever use of size to simplify code! What would the code look like if size was not available?

brzozan (changes requested):
Very, very good pull request!

Code is of high quality and is very well tested. Great job!

I added a few minor comments.

narciuch/assignment1/src/Anagram.java
narciuch: Completely agree, will change this to CHARS_TO_REMOVE

narciuch/assignment1/src/Anagram.java
narciuch: Not quite sure what you mean here. In isAnagramSentence(), preProcessSentence() is called initially on both sentences, which splits the sentences into words and sorts the characters within each word as well as sorting the entire word array. By sorting the entire word arrays, this ensures we are comparing the desired possible word anagrams.

I have added the example you have provided "abc def" and "def abc" to the test suite and this test passes.
I have also added a comment to preProcessSentence() to explain this, hopefully this helps.

narciuch/assignment1/src/Anagram.java
narciuch: Have added this.

narciuch/assignment1/src/Anagram.java
narciuch: Please see comment above. This is where I ensure the words in the anagram sentence can be in a different order.

narciuch/assignment1/src/SinglyLinkedList.java
narciuch: Would be possible to traverse the list once every time this code is called, but the extra time to do this does not outweigh the small amount of space required to simply keep track of the size. 

As you mentioned in the group session, another possibility would have been to use a moving window of size k with 2 pointers.

narciuch/assignment1/src/Anagram.java
brzozan: Makes sense. Turns out that I figured out your clever optimization only partially :) That's why comments in such places (where something clever/non-obvious is done) are so important. Thanks for clarification!

brzozan (approved):
LGTM. Please merge your PR. Good job!

PR #7, opened 2018-05-22
Stelaval Assignment 1

stelaval:


stelaval/assignment1/Anagram.java
```diff
@@ -0,0 +1,95 @@
...
+	static boolean isWordAnagram(String s1, String s2, boolean isCaseSensitive) {
+		if (isCaseSensitive) {
+			return isAnagram(s1, s2);
+		} else {
+			return isAnagram(s1.toLowerCase(), s2.toLowerCase());
+		}
+	}
+
+	/**
+	 * Determines if one sentence is an anagram of another sentence
```
brzozan, line 47: It would be good to add more details, so that if someone decides to use your code in their project, they immediately know what does it mean that two sentences are anagrams of each other.

stelaval/assignment1/Anagram.java
```diff
@@ -0,0 +1,95 @@
...
+	/**
+	 * Determines if one string is an anagram of another string
+	 * 
+	 * @param s1 first string
+	 * 
+	 * @param s2 second string
+	 * 
+	 * @return true if s1 and s2 are anagrams of each other
+	 */
+	static boolean isAnagram(String s1, String s2) {
```
brzozan: API methods should be public. Otherwise they are available only from within your package, so nobody else can use it.

stelaval/assignment1/Anagram.java
```diff
@@ -0,0 +1,95 @@
...
+
+import java.util.Arrays;
+
+public class Anagram {
+
+	/**
+	 * Determines if one string is an anagram of another string
+	 * 
+	 * @param s1 first string
+	 *
```
brzozan: No need to put empty line between params.

stelaval/assignment1/Anagram.java
```diff
@@ -0,0 +1,95 @@
...
+			return false;
+		} else {
+
+			boolean[] isAnagramFound = new boolean[sentence1.length];
+
+			for (int i = 0; i < sentence1.length; i++) {
+
+				for (int j = 0; j < sentence2.length; j++) {
+
+					if (isWordAnagram(sentence1[i], sentence2[j], false)) {
```
brzozan, line 72: No need to call the method with `isCaseSensitive` param. Just `isWordAnagram(sentence1[i], sententce2[j])`.

stelaval/assignment1/Anagram.java
```diff
@@ -0,0 +1,95 @@
...
+
+		if (sentence1.length != sentence2.length) {
+			return false;
+		} else {
+
+			boolean[] isAnagramFound = new boolean[sentence1.length];
+
+			for (int i = 0; i < sentence1.length; i++) {
+
+				for (int j = 0; j < sentence2.length; j++) {
```
brzozan: This algorithm works if one takes assignment literally - sentences are anagrams if every word from first one is an anagram of any word from the second. This means that "abc abc" and "abc def" are anagrams.

What if the definition was narrowed down so that each word from each sentence can be used only once, i.e. "abc abc" and "abc def" are not anagrams, but "abc def" and "abc fed" are. And similarly "abc abc" and "abc abc abc" were not anagram.

How would that change algorithm?

stelaval/assignment1/SinglyLinkedList.java
```diff
@@ -0,0 +1,95 @@
...
+		int length = 0;
+		Node<T> currentNode = getHead();
+
+		while (currentNode != null) {
+			currentNode = currentNode.getNext();
+			length++;
+		}
+
+		if (length < k) {
+			try {
```
brzozan: No need to catch it here. Either declare exception in the method (`public T getKthToLastElement(int k) throws Exception`) and catch it in `main`, or use exception that inherits from `RuntimeException` which does not need to be declared.

In this case the best approach would be to throw `IllegalArgumentException` (it is a `RuntimeException`, so no need to declare it in method).

stelaval/assignment1/SinglyLinkedList.java
```diff
@@ -0,0 +1,95 @@
...
+				throw new Exception("Index k is more than the length of the linked list");
+			} catch (Exception e) {
+				e.printStackTrace();
+			}
+		}
+
+		currentNode = getHead();
+
+		for (int i = 1; i < length - k + 1; i++)
```
brzozan, line 33: This requires traversing the list twice - once to calculate length, then second time to find the element. How could the algorithm be changed (without storing length in a field!) to traverse list only once?

Imagine that someone wants 100th element from the end of a list with 10000000000 elements.

stelaval/assignment1/SinglyLinkedList.java
```diff
@@ -0,0 +1,95 @@
...
+		head = newNode;
+	}
+
+	private class Node<T> {
+
+		private T data;
+		private Node next;
+
+		private Node() {
+			this.data = data;
```
brzozan: This line really does not do anything and can be safely removed. It is basically `this.data = this.data`.

Actually the whole constructor could probably be removed since it is not used.

stelaval/assignment1/SinglyLinkedList.java
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		private Node(T data) {
```
brzozan: It could be either public or package private.

stelaval/assignment1/SinglyLinkedList.java
```diff
@@ -0,0 +1,95 @@
...
+	public void push(int data) {
+
+		Node<T> newNode = new Node(data);
+
+		newNode.next = head;
+
+		head = newNode;
+	}
+
+	private class Node<T> {
```
brzozan: It is usually a good idea to make such internal classes `private static` rather than just `private` - this saves a few bytes with are otherwise used to store reference to `SinglyLinkedList`.

stelaval/assignment1/SinglyLinkedList.java
```diff
@@ -0,0 +1,95 @@
...
+
+		private Node(T data) {
+			this.data = data;
+		}
+
+		public T getData() {
+			return data;
+		}
+
+		private void setData(T data) {
```
brzozan: `setData` is not used, so it is better to remove it.

stelaval/assignment1/SinglyLinkedList.java
```diff
@@ -0,0 +1,95 @@
...
+		Node<T> newNode = new Node(data);
+
+		newNode.next = head;
+
+		head = newNode;
+	}
+
+	private class Node<T> {
+
+		private T data;
```
brzozan: This field is assigned only from constructor (`setData` is not used anywhere), so it would be good to make it `final`. This lets Java optimize code a bit better and avoids potential problems in the future when someone decides to start modifying this field while other parts of code do not expect it.

stelaval/assignment1/SinglyLinkedList.java
```diff
@@ -0,0 +1,95 @@
...
+
+		private void setData(T data) {
+			this.data = data;
+		}
+
+		public Node getNext() {
+			return next;
+		}
+
+		private void setNext() {
```
brzozan: `setNext` is not used anywhere.

brzozan (changes requested):
Very nice and clean PR! Thank you!

General comment: please follow Google coding style (which is almost identical to the style that Sun and later Oracle recommended for Java code).
That means two spaces used for indentation rather than tab character. It is a minor thing, but tabs by default expand to 8 spaces which makes code take a lot of horizontal space and forces eyes to work harder while reading it.

I added a few more detailed comments to the code.

stelaval/assignment1/Anagram.java
stelaval: Will have that in mind

stelaval/assignment1/Anagram.java
stelaval: I will change it

stelaval/assignment1/Anagram.java
stelaval: By default, isAnagram is case sensitive. As long as sentences start with capital letter, I wanted to use a method that is case insensitive which is isWordAnagram(sentence1[i], sentence2[j], false)

stelaval/assignment1/SinglyLinkedList.java
brzozan: As per discussion during our meeting on Friday, feel free to ignore this comment.

stelaval/assignment1/SinglyLinkedList.java
brzozan: It is better to just remove unused code. This makes code smaller and you don't have to worry about maintenance of code that is not necessary.

brzozan (changes requested):
Thank you for doing some of the requested changes! Could you please consider the remaining comments too?

stelaval:
Thank you very much for the review. All the requested changes are done now.

brzozan (approved):
Thanks for doing the changes!

The formatting is not as per Google coding style, but it is not a blocking issue and other than that code looks good to me.

PR #8, opened 2018-05-22
Submit Assignment 1

yujo-t:


uro98/assignment1/Anagram.java
```diff
@@ -0,0 +1,72 @@
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+
+public class Anagram {
+
+	public static Boolean wordAnagram(String a, String b, Boolean sensitive) {
```
brzozan: No need to use `Boolean`, just `boolean`. `Boolean` with capital B is useful only in collections (for example `Map<String, Boolean>`) or when you need three states (`true`, `false` and `null`, in which case it is better to use `Optional<Boolean>` instead of `Boolean` anyway).

Also it would be better to name last parameter `caseSensitive` to make it clean what kind of sensitivity it represents.

uro98/assignment1/Anagram.java
```diff
@@ -0,0 +1,72 @@
...
+
+		// Fill histograms
+		for (int i = 0; i < a.length(); i++) {
+			if (hisA.containsKey(a.charAt(i))) {
+				hisA.put(a.charAt(i), hisA.get(a.charAt(i)) + 1);
+			} else {
+				hisA.put(a.charAt(i), 1);
+			}
+
+			if (hisB.containsKey(b.charAt(i))) {
```
brzozan: This block is very similar to the previous one. Could you move it to a helper method to avoid code duplication?

uro98/assignment1/Anagram.java
```diff
@@ -0,0 +1,72 @@
...
+		HashMap<Character, Integer> hisA = new HashMap<Character, Integer>();
+		HashMap<Character, Integer> hisB = new HashMap<Character, Integer>();
+
+		if (a.length() != b.length()) {
+			return false;
+		}
+
+		// Fill histograms
+		for (int i = 0; i < a.length(); i++) {
+			if (hisA.containsKey(a.charAt(i))) {
```
brzozan: Optional: this whole `if` could be replaced with something like: `hisA.put(a.charAt(i), hisA.getOrDefault(a.charAt(i), 0) + 1);`.

uro98/assignment1/Anagram.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class Anagram {
+
+	public static Boolean wordAnagram(String a, String b, Boolean sensitive) {
+		if (!sensitive) {
+			a = a.toLowerCase();
+			b = b.toLowerCase();
+		}
+
+		HashMap<Character, Integer> hisA = new HashMap<Character, Integer>();
```
brzozan: Optional: variable types can be just `Map<Character, Integer>` since in code below you do not actually care about what kind of map is being used.

uro98/assignment1/Anagram.java
```diff
@@ -0,0 +1,72 @@
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+
+public class Anagram {
```
brzozan: Could you please add javadoc comments to public methods? API should have some documentation that would make it easy for people not familiar with your class to use it in their code.

uro98/assignment1/KToLast.java
```diff
@@ -0,0 +1,58 @@
...
+
+	Node first;
+
+	public SinglyLinkedList(Object obj) {
+		first = new Node(obj);
+	}
+
+	class Node {
+
+		Object data;
```
brzozan: Just storing `Object` is not very convenient for users of the class as they have to remember what data they put inside and then cast it back to their specific type. Could you change code to use generics?

uro98/assignment1/KToLast.java
```diff
@@ -0,0 +1,58 @@
...
+		SinglyLinkedList.Node third = list.new Node(3);
+
+		list.first.next = second;
+		second.next = third;
+
+		System.out.println(kToLastObject(list, 1));
+	}
+}
+
+class SinglyLinkedList {
```
brzozan: To make code reusable it would be better if `SinglyLinkedList` was a public class in a separate file.

Also: please add javadoc.

uro98/assignment1/KToLast.java
```diff
@@ -0,0 +1,58 @@
...
+
+	Node first;
+
+	public SinglyLinkedList(Object obj) {
```
brzozan: It would be useful to be able to create empty list (without any objects) too...

uro98/assignment1/KToLast.java
```diff
@@ -0,0 +1,58 @@
...
+		first = new Node(obj);
+	}
+
+	class Node {
```
brzozan: `Node` should not be exposed to outside world, so it should be `private`. It would be good to make it also as `static`, since it does not  access `SinglyLinkedList`'s fields.

Instead of making this class visible and manually modifying `next`, please add `add` method in `SinglyLinkedList` that would add new element to the list.

uro98/assignment1/KToLast.java
```diff
@@ -0,0 +1,58 @@
+public class KToLast {
+
+	public static Object kToLastObject(SinglyLinkedList list, int k) {
```
brzozan: It would be much better if this method was part of the `SinglyLinkedList` rather than separate static one.

uro98/assignment1/KToLast.java
```diff
@@ -0,0 +1,58 @@
...
+			length++;
+			node = node.next;
+		}
+
+		int target = length - k - 1;
+		int index = 0;
+		node = list.first;
+
+		// Get target node
+		while (index != target) {
```
brzozan: This requires list to be traversed twice (once to find length and once to find the actual element). How could the code be changed to traverse it only once?

Imagine that someone wants to find 100th element from the end of a list that has 10000000000 elements.

brzozan (changes requested):
Very nice and clean PR. Thank you!

First, global stylistic comment: as per Google Java style [1], two spaces should be used for indentation. It is a minor thing, but tabs are expanded to 8 spaces by default which makes code take a lot of horizontal space.

Second global comment: please add javadoc to public methods. Someone (or even you 5 years from now) might want to use your class in their code, so it would be good to give them easy way to know what it does without the need to go through code.

I also added some more detailed comments in code.

[1] Style guide: https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation

uro98/assignment1/Anagram.java
yujo-t: I am now using `getOrDefault` as you suggested which reduced the two blocks to two lines, is it okay now if I don't create a helper method for this?

yujo-t:
Thank you for the feedback Michal, I have changed the code accordingly.

uro98/assignment1/Anagram.java
```diff
@@ -1,72 +1,80 @@
...
+      hisB.put(b.charAt(i), hisB.getOrDefault(b.charAt(i), 0) + 1);
+    }
 
-			if (hisB.containsKey(b.charAt(i))) {
-				hisB.put(b.charAt(i), hisB.get(b.charAt(i)) + 1);
-			} else {
-				hisB.put(b.charAt(i), 1);
-			}
-		}
+    if (hisA.equals(hisB)) {
```
brzozan: Just `return his!.equals(hisB);`.

brzozan (approved):
LGTM. I added one small comment, but other than that PR is in good enough shape to be merged. Thank you!

PR #9, opened 2018-06-01
Assignment 2 - YakimS

YakimS:


YakimS/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,231 @@
+package CodeUAss1Package;
+
+import javax.management.openmbean.KeyAlreadyExistsException;
```
brzozan: It is better not to depend on such complex packages just for exception. Better use something from `java.lang`, like `IllegalArgumentException`.

YakimS/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,231 @@
+package CodeUAss1Package;
+
+import javax.management.openmbean.KeyAlreadyExistsException;
+import java.util.*;
```
brzozan: Please avoid importing via asterisk.

YakimS/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * This class contains the root to a binary tree that can have only
+ * one node for each possible key.
+ * @param <T> is the type of data stored in the tree nodes value
+ */
+public class BinaryTree<T> {
+
+    private Node root;
+    private Set<Integer> keySet;
```
brzozan: The whole tree has generic parameter `<T>`, so `keySet` should probably also use it.

YakimS/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,231 @@
...
+    public BinaryTree() {
+        keySet = new HashSet<>();
+    }
+
+    public void printBinaryTree()
+    {
+        printBinaryTree(this.root,0);
+    }
+    private void printBinaryTree(Node root, int level){
+        if(root==null)
```
brzozan: Please avoid `if`s, `for`s etc without curly braces. It is hard to read and error-prone.

YakimS/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,231 @@
...
+        printBinaryTree(root.getLeftSon(), level+1);
+    }
+
+    /**
+     * This method add a node to a random valid place in the binary tree
+     * @param value - value of the added node
+     * @param key - key of the added node
+     * @throws KeyAlreadyExistsException if there is already a node with key as given in the parameter
+     */
+    public void add(T value, int key) throws KeyAlreadyExistsException
```
brzozan: Key is not really needed, since it is not intended to be a map. Better to just use value as the key.

YakimS/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,231 @@
...
+            return;
+        }
+
+        /*
+        *   Moving down the tree, while choosing the path down (through right or left son)
+        *  randomly, till the current node is null, and then assign to the added node
+        *  to that null current node
+        */
+        Node currentNode = root;
+        Random random = new Random();
```
brzozan: Using a randomized data structure is an interesting idea, but it makes it very hard to test and to reason about. Unit tests are not required for this assignment, but if they were, it would be nearly impossible to check if they are correct, and since you use `Random` without seed, the results you will get are very likely to be different each time you run the program.

YakimS/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,231 @@
...
+     * @param key1 - first key to find common ancestor
+     * @param key2 - second key to find common ancestor
+     * @return lowest common ancestor of the two given keys. If the keys are equal than
+     *          the function will return the key number.
+     * @throws KeyNotFoundException if one or both of the keys are not in the tree
+     */
+    public int getLowestCommonAncestor(int key1, int key2) throws KeyNotFoundException {
+
+        // create paths to the two keys given
+        List<Integer> firstKeyPath = getKeyPath(key1, root, new LinkedList<>());
```
brzozan: The assignment explicitly asked to avoid storing nodes in additional data structures. While technically you are storing keys instead of nodes, in practice it is the same. Could you change the code to avoid that?

brzozan (changes requested):
Thank you for the PR!

Top-level comment: please format code consistently - braces placed always in the same way, spaces used the same way between keywords and parentheses etc.
Currently it is quite difficult to read, because formatting varies between methods and statements.

I added a few more detailed comments in code.

YakimS/assignment2/BinaryTree.java
YakimS: The generic parameter is referred to the type of data the tree stores in his nodes, and I choose to store the key as integers. Can you please explain why should I use the keys of the nodes to be the same type as the value they reference?

YakimS: You've answered my question at the next change request.

YakimS:
I have assigned the requested changes.
Thank you very much for the review!

YakimS/assignment2/BinaryTree.java
```diff
@@ -124,108 +71,83 @@ else if(printAncestors(key,root.getLeftSon()) ||
...
+        TreeNode<T> nodeInPathOfVal1 = findNodeByVal(value1, root);
+        TreeNode<T> nodeInPathOfVal2 = findNodeByVal(value2, root);
 
-        if(firstKeyPath == null || secKeyPath == null)
+        if(nodeInPathOfVal1 == null || nodeInPathOfVal2 == null)
         {
-            throw new KeyNotFoundException();
+            throw new IllegalArgumentException("Value not found");
+        }
+        else if(nodeInPathOfVal1.equals(nodeInPathOfVal2))
```
brzozan: Optional: At this point you are operating only on nodes that are already in the tree and do not change, so it is OK to just use `==` for comparison.

YakimS/assignment2/BinaryTree.java
```diff
@@ -124,108 +71,83 @@ else if(printAncestors(key,root.getLeftSon()) ||
...
-            int nextInFirstPath = (int) IteratorFirst.next();
-            int nextInSecondPath = (int) IteratorSecond.next();
-            if(nextInFirstPath == nextInSecondPath)
+        while (nodeInPathOfVal1.getLevel() != 0 && nodeInPathOfVal2.getLevel()!=0)
+        {
+            if(nodeInPathOfVal1.equals(nodeInPathOfVal2))
+            {
+                return nodeInPathOfVal1.getValue();
+            }
+            else if(nodeInPathOfVal1.getLevel() < nodeInPathOfVal2.getLevel())
```
brzozan, line 107: Optional: No need for `else` here. Having two separate ifs (one for node equality, one for level calculations) will be a bit easier to read, since they are logically separate.

YakimS/assignment2/BinaryTree.java
```diff
@@ -124,108 +71,83 @@ else if(printAncestors(key,root.getLeftSon()) ||
...
-    private LinkedList<Integer> getKeyPath(int key, Node root, LinkedList<Integer> linkedList)
+
+    private TreeNode<T> findNodeByVal(T value1, TreeNode<T> root)
     {
         if(root == null)
         {
             return null;
         }
-        if(root.getKey() == key)
+        else if(root.getValue() == value1)
```
brzozan: This should be `.equals`.  Two `Integer` (or `String`, or ...) objects can carry the same value, but not be equal when compared via `==`.

brzozan (approved):
Thank you for applying requested changes!

I have a few optional comments and one non-optional (non-optional, because it is a bug).

I am approving this PR, but before merging please fix the non-optional comment. Thank you!

PR #10, opened 2018-06-01
Narciuch 2

narciuch:


narciuch/assignment2/src/Node.java
```diff
@@ -0,0 +1,29 @@
+public class Node<T> {
+  private T value;
```
brzozan: Value does not change over lifetime of the node, so it would be good to make it `final`.

narciuch/assignment2/src/TreeAncestors.java
```diff
@@ -0,0 +1,135 @@
...
+   * Used to find the path from the root of a tree to a particular node
+   * (i.e. the nodes ancestors)
+   *
+   * @param key : The value of the desired node
+   * @param node : Tree to search through
+   * @param path : Indicates the visited nodes in the path from the original root directly to
+   *             the current 'node'
+   * @return true if the node with value 'key' was found in the tree 'node'
+   */
+  public boolean getAncestorsHelper(T key, Node node, List<Node> path) {
```
brzozan: Helper methods are usually not meant to be used by clients of the class and as such should be made private.

narciuch/assignment2/src/TreeAncestors.java
```diff
@@ -0,0 +1,135 @@
...
+    boolean right = false;
+    if (node.getLeft() != null) {
+      path.add(node);
+      left = getAncestorsHelper(key, node.getLeft(), path);
+    }
+    if (node.getRight() != null) {
+      path.add(node);
+      right = getAncestorsHelper(key, node.getRight(), path);
+    }
+    if (path.size() != 0 && !left && !right) {
```
brzozan, line 64: Current code does not seem to enforce uniqueness of keys and there is no comment that would suggest that it is a requirement, so it is possible that node with given value is present in both left and right subtree.
What will happen in such case?

Comment is about the method in general, not necessarily line 58 :)

narciuch/assignment2/src/TreeAncestors.java
```diff
@@ -0,0 +1,135 @@
...
+   *
+   * @param n1    : First node to look for
+   * @param n2    : Second node to look for
+   * @param node  : Root of the tree to search for n1 and n2 in
+   * @param found : How many of n1 and n2 have already been found
+   * @return 0 if neither n1 nor n2 are in the 'node' tree
+   * 1 if either n1 or n2 are in the 'node' tree
+   * 2 if both n1 and n2 are in the 'node' tree
+   */
+  public int nrNodesFound(Node<T> n1, Node<T> n2, Node<T> node, int found) {
```
brzozan: This seems to be a helper method, so it should be private.

narciuch/assignment2/test/TreeAncestorsTest.java
```diff
@@ -0,0 +1,96 @@
...
+  @Test
+  public void correctlyIdentifiesAncestors() {
+    Node three = new Node(3);
+    Node four = new Node(4);
+    Node two = new Node(2);
+    Node five = new Node(5);
+    Node eight = new Node(8);
+    Node one = new Node(1);
+    Node six = new Node(6);
+    seven.setLeft(three);
```
brzozan: This way of building tree seems quite inconvenient. It would be much easier if `Node` provided methods like:
```
/** Adds node with given value as left child of current node and returns newly created child. */
public Node<T> addLeftChild(T value) {...}
```

narciuch/assignment2/test/TreeAncestorsTest.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @Test
+  public void noAncestorsIfKeyNotInTree() {
+    assertTrue(ancestorFinder.getAncestors(44, getTestTree()).size() == 0);
+  }
+
+  @Test
+  public void correctlyIdentifiesAncestors() {
+    Node three = new Node(3);
```
brzozan: Generic parameters missing, which uses compile-time type information and thus makes code more error-prone.

narciuch/assignment2/src/TreeAncestors.java
```diff
@@ -0,0 +1,135 @@
...
+      path.add(node);
+      left = getAncestorsHelper(key, node.getLeft(), path);
+    }
+    if (node.getRight() != null) {
+      path.add(node);
+      right = getAncestorsHelper(key, node.getRight(), path);
+    }
+    if (path.size() != 0 && !left && !right) {
+      node = path.get(path.size() - 1); //Take most recently added node
+      path.remove(path.size() - 1);
```
brzozan, line 66: Optional: would it be possible to reorganize code so that elements are added to the path only when necessary rather than always added and then often removed?

brzozan (changes requested):
Thank you for really good PR!

I added a few comments to the code.

narciuch/assignment2/src/Node.java
narciuch: Agreed

narciuch/assignment2/src/TreeAncestors.java
narciuch: Good point

narciuch/assignment2/src/TreeAncestors.java
narciuch: Agreed

narciuch/assignment2/test/TreeAncestorsTest.java
narciuch: Yes, this is a much nicer solution.

narciuch/assignment2/src/TreeAncestors.java
narciuch: Very good point.
Have added a test and comment to getAncestors() about the uniqueness of keys.

As for the comment on line 58, it seemed to be more of an implementation detail, explaining exactly what is happening within that part of the code, irrelevant to the rest of the function. 
I'm not quite sure what you mean here.

narciuch/assignment2/test/TreeAncestorsTest.java
narciuch: Good point

brzozan (approved):
LGTM. Thank you!

PR #11, opened 2018-06-05
Create BinaryTree.java

VolhaAnishchanka:
Implemented methods for searching for lca and printing all ancestors of a node in a binary tree.


HelgaShiryaeva/assignment2/OutOfTreeException.java
```diff
@@ -0,0 +1,14 @@
+package assignment2;
+
+/**
+ * Created by Helga on 6/2/2018.
+ */
+public class OutOfTreeException extends Exception {
```
brzozan: Dedicated exception is not really needed. It would be better to just throw either `NoSuchElementException` or `IllegalArgumentException`, since this seems to be what you try to communicate using `OutOfTreeException`.

This would have two advantages over dedicated checked (i.e. inheriting from `Exception` rather than from `RuntimeException`) exception:
- It would be consistent with the standard library.
- It would not force users of your class to wrap each use of it with `try { ... } catch (...) {...}` block, which is inconvenient and actually often leads to just catching all exceptions and ignoring them.

Checked exceptions have their uses, but should be used sparingly.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * All the nodes of a tree should be inserted into a nodes set
+     * in order to avoid duplicates. A node which is not in the
+     * set of nodes is not considered to be a part of a binary tree.
+     * @param newNode is a node to be inserted in the nodes set
+     * @throws IllegalArgumentException if node is already in a binary tree or if
+     * the node with such value is already exists in the tree
+     */
+    public void addNode(Node<T> newNode) {
```
brzozan: This method puts the whole burden of actually constructing the tree (setting left and right) on the user, which is error-prone and quite inconvenient.

It would be much easier to use if it looked something like:
```
public enum Side {
  LEFT,
  RIGHT
}

/**
 * ...
 * @param newValue value to be inserted to the tree.
 * @param parent parent of the new node.
 * @param side on which side of the parent new node should be added.
 * @returns newly created node.
 */
public Node<T> addNode(T newValue, Node<T> parent, Side side);
```
Of course that would need to handle additional case - what to do if user tries to insert node on the if parent already has child on requested side. It could either throw or just insert new child between parent and old child.

Even more convenient API would be to hide Node class completely and operate only on values, especially that this class already assumes that values are unique.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,104 @@
...
+            printAncestorsRecursive(node, root);
+            System.out.println();
+        } else {
+            throw new OutOfTreeException(node);
+        }
+    }
+
+    private Node<T> printAncestorsRecursive(Node<T> node, Node<T> currentNode) throws OutOfTreeException {
+        if (currentNode != null) {
+            if (!nodes.contains(currentNode)) {
```
brzozan: This check should not be necessary if the tree was well hermetized, i.e. users of the class were not allowed to modify children of the node.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,104 @@
...
+                throw new OutOfTreeException(currentNode);
+            }
+            if (currentNode.getValue().equals(node.getValue())) {
+                return currentNode;
+            }
+            Node<T> left = printAncestorsRecursive(node, currentNode.getLeft());
```
brzozan: Optional: `printAncestorsRecursive` could just return `boolean` indicating whether node was found in given subtree.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,104 @@
...
+    public void printAncestors(Node<T> node) throws OutOfTreeException {
+        if (nodes.contains(node)) {
+            printAncestorsRecursive(node, root);
+            System.out.println();
+        } else {
+            throw new OutOfTreeException(node);
+        }
+    }
+
+    private Node<T> printAncestorsRecursive(Node<T> node, Node<T> currentNode) throws OutOfTreeException {
```
brzozan: It would be nice to have a comment explaining at least what the return value is. It would make understanding code a bit easier.

HelgaShiryaeva/assignment2/Node.java
```diff
@@ -0,0 +1,36 @@
...
+ * @param <T> is a type of value of the node
+ */
+public class Node <T> {
+    private T value;
+    private Node left;
+    private Node right;
+
+    public Node(T value) {
+        this.value = value;
+        left = right = null;
```
brzozan: Not really needed. Fields are null by default.

HelgaShiryaeva/assignment2/Node.java
```diff
@@ -0,0 +1,36 @@
+package assignment2;
+
+/**
+ * This class represents a node of a binary tree
+ * @param <T> is a type of value of the node
+ */
+public class Node <T> {
+    private T value;
```
brzozan: Value does not change over lifetime of the node, so it's better to make it explicitly `final`.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,104 @@
...
+        if (currentNode != null) {
+            if (!nodes.contains(currentNode)) {
+                throw new OutOfTreeException(currentNode);
+            }
+            if (currentNode.getValue().equals(node.getValue())) {
```
brzozan: Optional, but good practice in general: currently you do not prevent nodes with `null` value from being created and added to the tree, which will result in `NullPointerException` here. Either node (or tree) should throw exception if `null` value is set or whenever you call methods on value you need to check if it is not null.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,104 @@
...
+                return currentNode;
+            }
+            Node<T> left = printAncestorsRecursive(node, currentNode.getLeft());
+            Node<T> right = printAncestorsRecursive(node, currentNode.getRight());
```
brzozan: There is an optimization opportunity here - since you already enforce that values in the tree are unique, there is no need to traverse right subtree if element was found in the left one.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,104 @@
...
+            throw new OutOfTreeException("Node(s} is(are) not present in a binary tree");
+        }
+        return getLcaRecursive(root, first, second);
+    }
+
+    private Node<T> getLcaRecursive(Node<T> current, Node<T> first, Node<T> second) throws OutOfTreeException {
+        if (current == null) {
+            return null;
+        }
+        if(!nodes.contains(current)) {
```
brzozan: Should not be necessary.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,104 @@
...
+         return null;
+    }
+
+    /**
+     * Finds the lowest common ansestor of two nodes of a binary tree
+     * @return node which is the lowest common ancestor of two nodes
+     * @throws OutOfTreeException if either fisrt or second or both node(s) are not a part
+     * of a binary tree
+     */
+    public Node<T> getLca(Node<T> first, Node<T> second) throws OutOfTreeException {
```
brzozan: It would be easier to use if public API used values instead of nodes. For example current implementation will not work on your test data if I call it with `getLca(new Node(7), new Node(3))`, even though both 7 and 3 are in the tree.

HelgaShiryaeva/assignment2/Node.java
```diff
@@ -0,0 +1,36 @@
+package assignment2;
+
+/**
+ * This class represents a node of a binary tree
+ * @param <T> is a type of value of the node
+ */
+public class Node <T> {
```
brzozan: Optional: `Node` is closely related to `BinaryTree` class, so it could be an inner class of `BinaryTree`.

HelgaShiryaeva/assignment2/Node.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public T getValue() {
+        return value;
+    }
+
+    public Node<T> getLeft() {
+        return left;
+    }
+
+    public void setLeft(Node left) {
```
brzozan: Optional: making such methods publicly visible is quite dangerous as it can result in client code breaking consistency of your tree, for example by adding node correctly to the tree, but then calling `setLeft(new Node(-7))` on it.
You worked around it by adding checks that all nodes are present in set, but it would be much safer to hide these methods by making them either package private (just remove `public`) or completely private (possible if Node class inside `BinaryTree` and logic of setting children is hidden inside `addNode` method).

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private Node<T> getLcaRecursive(Node<T> current, Node<T> first, Node<T> second) throws OutOfTreeException {
+        if (current == null) {
+            return null;
+        }
+        if(!nodes.contains(current)) {
+            throw new OutOfTreeException(current);
+        }
+        if (current == first || current == second) {
```
brzozan: It would be better to compare values rather than nodes.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,104 @@
+package assignment2;
+
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * This class represents a Binary tree data structure
```
brzozan: It would be good to also mention assumptions that this class makes, i.e. that values in tree are unique.

brzozan (changes requested):
Thanks you for this PR! Well done!

I do not have algorithmic comments as the code seems to be doing what it is supposed to do, but I added some comments about code structure and interface.

Those comments can be summarized as: try to hide as much of implementation details of your class from the user, exposing as simple public interface as possible.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -1,104 +1,166 @@
...
 
 import java.util.HashSet;
-import java.util.Optional;
+import java.util.NoSuchElementException;
 import java.util.Set;
 
 /**
  * This class represents a Binary tree data structure
  * @param <T> is a type of values stored in nodes of a tree
+ * All values stored in a binary tree are unique
```
brzozan: This should go before params list. Typical structure of comment is:

- Summary.
- More details and assumptions.
- Params.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -1,104 +1,166 @@
...
-        }
-        nodes.add(newNode);
+     * This method adds a new node in a binary tree
+     * If the parent of a new node already has a child on
+     * the requested side an old child would be replaced by a new one
+     * @param value is a value of a new node
+     * @param parent is a parent of the node
+     * @param side on which side of parent a new node should be added
+     * */
+    public void add(T value, T parent, Side side) {
```
brzozan, line 31: Optional: if the class assumes that values are unique, it would be good to enforce it at the time when new element is added.

HelgaShiryaeva/assignment2/BinaryTree.java
```diff
@@ -1,104 +1,166 @@
...
-    private Node<T> getLcaRecursive(Node<T> current, Node<T> first, Node<T> second) throws OutOfTreeException {
+    private Node getLcaRecursive(Node current, T first, T second) {
         if (current == null) {
             return null;
         }
-        if(!nodes.contains(current)) {
-            throw new OutOfTreeException(current);
-        }
-        if (current == first || current == second) {
+        if (current.getValue() == first || current.getValue() == second) {
```
brzozan: Comparison of values should be done with `equals()` rather than `==`. For example two `Integer`s can carry the same value but not be equal when compared with `==`.

brzozan (approved):
Thank you for making requested changes!

I will approve this PR, but before merging please see the comment about equals() vs == and change code.
Other comments are optional.

VolhaAnishchanka:
Thank you so much !
You found a real bugs in my class, code reviews are a really cool thing !

PR #12, opened 2018-06-05
Submit Assignment 2

yujo-t:


uro98/assignment2/Node.java
```diff
@@ -0,0 +1,19 @@
+/**
+ * Represents a generic node in a generic binary tree.
+ */
+public class Node<T> {
+  
+  T data;
```
brzozan: It would be better to make it at least `final` to avoid accidental modification. Ideally field should be private and have getter method.

uro98/assignment2/Node.java
```diff
@@ -0,0 +1,19 @@
...
+  Node<T> left;
+  Node<T> right;
+  Node<T> parent;
+  int depth;
+  
+  public Node(T data, int depth) {
+    this.data = data;
+    left = null;
```
brzozan: It is not necessary to initialize fields with `null`, since it is the default value they have.

uro98/assignment2/LowestCommonAncestor.java
```diff
@@ -0,0 +1,65 @@
...
+   * @return      the value of the lowest common ancestor
+   */
+  public static <T> T lowestCommonAncestor(BinaryTree<T> tree, Node<T> node1, Node<T> node2) {
+    
+    int depth1 = node1.depth;
+    int depth2 = node2.depth;
+    Node<T> temp = node2;
+    
+    // For each ancestor of node1, look for an identical ancestor of node2
+    for (int i = depth1; i > 0; i--) {
```
brzozan: Is there any other way to know when you reached end of the tree without using depth?

uro98/assignment2/LowestCommonAncestor.java
```diff
@@ -0,0 +1,65 @@
...
+      node1 = node1.parent;
+      node2 = temp;
```
brzozan: It would be a little bit cleaner if you did it the other way round and used `temp` to walk through the tree, every time resetting it to `temp = node2` here.

uro98/assignment2/PrintAncestors.java
```diff
@@ -0,0 +1,84 @@
...
+  
+  /**
+   * Print the ancestors of the node with the given key in a binary tree as a list.
+   * Uses left-to-right depth first search.
+   * An empty list is printed if the node is not found.
+   * @param tree  a generic binary tree
+   * @param key   the target value of a node
+   */
+  public static <T> void printAncestors(BinaryTree<T> tree, T key) {
+    LinkedList<T> ancestors = new LinkedList<T>();
```
brzozan: Optional, but good practice in general: unless there is a really good reason, it is better to use `ArrayList`, since it performs a bit better. There is a good reason to use `LinkedList` for `frontier`, but not really for `ancestors`.

uro98/assignment2/PrintAncestors.java
```diff
@@ -0,0 +1,84 @@
...
+    LinkedList<Node<T>> frontier = new LinkedList<Node<T>>();
```
brzozan: Optional: type of variable could be `Deque<Node<T>>` to make it more explicit that it is used as a queue.

uro98/assignment2/PrintAncestors.java
```diff
@@ -0,0 +1,84 @@
...
+          for (int i = 0; i < depthDifference + 1; i++) {
+            ancestors.pop();
+          }
+        }
+      }
+    }
+    
+    // If the node is not found, print an empty list
+    ancestors.clear();
+    printList(ancestors);
```
brzozan: Optional: instead of sequence `ancestors.clear(); printList(ancestors);` you could just skip clearing and do `printList(Collections.emptyList())`, which would make it explicit what is happening.

uro98/assignment2/PrintAncestors.java
```diff
@@ -0,0 +1,84 @@
...
+  public static <T> void printList(LinkedList<T> list) {
+    System.out.println(Arrays.toString(list.toArray()));
+  }
+
+  public static void main(String[] args) {
+    BinaryTree<Integer> tree = new BinaryTree<Integer>();
+    
+    tree.addRoot(7);
+    
+    Node<Integer> three = new Node<Integer>(3, 1);
```
brzozan: Explicitly setting depth at the time when node is constructed and then manually setting up parent and children is not too convenient and very error-prone. It would be better if all this logic was hidden inside tree or node class.
One of possible solutions could be adding two methods in `Node`:
```
/** Adds node with given value as left child of current node and returns newly created child. */
public Node<T> addLeftChild(T value) {...}

/** Adds node with given value as left child of current node and returns newly created child. */
public Node<T> addRightChild(T value) {...}
```
Alternatively: a single `addChild` method that takes enum `Side` with values `LEFT` and `RIGHT`.

Then depth could be calculated automatically and if someone tried to use your tree in their code, they would not need to worry about manually constructing whole tree.

It would also let you make `left` and `right` fields private and provide getter methods, which would make your class much safer.

uro98/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,20 @@
+/**
+ * Represents a generic binary tree.
+ *
+ */
+public class BinaryTree<T> {
+  
+  Node<T> root;
```
brzozan: It would be safer if this field was private and a getter method was provided.

uro98/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,20 @@
...
+  
+  /**
+   * Set the root node to the value specified.
+   * @param data a generic value
+   */
+  public void addRoot(T data) {
+    Node<T> newNode = new Node<T>(data, 0);
+    
+    if (root == null) {
```
brzozan: Ideally this check should be done before creating node and if `root` is not `null`, something like `IllegalStateException` was thrown to signal error.

uro98/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public class BinaryTree<T> {
+  
+  Node<T> root;
+  
+  /**
+   * Set the root node to the value specified.
+   * @param data a generic value
+   */
+  public void addRoot(T data) {
```
brzozan: Optional: for convenience of use this method could return newly created `Node` so that users can easily use it to continue building the tree.

uro98/assignment2/PrintAncestors.java
```diff
@@ -0,0 +1,84 @@
...
+public class PrintAncestors {
+  
+  /**
+   * Print the ancestors of the node with the given key in a binary tree as a list.
+   * Uses left-to-right depth first search.
+   * An empty list is printed if the node is not found.
+   * @param tree  a generic binary tree
+   * @param key   the target value of a node
+   */
+  public static <T> void printAncestors(BinaryTree<T> tree, T key) {
```
brzozan: This method uses iterative implementation of DFS using queue with non-trivial logic for removal of ancestors based currently visited tree depth. It is absolutely not wrong, but is a bit more complicated than it could.

Optional: what would the code look like with recursive solution? Recursion is a very powerful tool and is used quite a lot in real life, so it is good to be familiar with it.

brzozan (changes requested):
Thank you for this PR! You did very good job!

Most of my comments are optional. Algorithms seem to be correct, even if a bit more complex than they could.

The thing that can be improved most is interface around building the tree. Currently it is not convenient to use and very error-prone.
Please see more details in comments.

uro98/assignment2/PrintAncestors.java
yujo-t: I'm using `frontier` as a stack though, should I still change it to `Deque<Node<T>>`?

brzozan: `Deque` is what defines those methods, so it would be a bit more clear, but the comment itself is optional, so feel free to leave it as is.

The idea behind my comment: type of variable can provide additional information about how you intend to use it. If the type is `LinkedList`, it tells me only that it is a general container with all the methods that it provides. If type is `Deque` it tells me that you plan to use it as queue or stack and makes it easier to spot situations where it might be misused (say: intended to use it as stack but then looking at random elements inside it via `get(index)`).

yujo-t: Right, that makes sense, thank you. I have now changed my DFS implementation to use recursion though so I've removed most of the code in `printAncestors()`.

brzozan (approved):
Thank you, LGTM!

PR #13, opened 2018-06-06
Assignment 2

stelaval:


stelaval/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,176 @@
...
+   * 
+   * @param key is a node which ancestors are printed
+   * 
+   * @throws Exception if given key is not present in a tree
+   */
+  public void printAncestors(Node<T> key) throws Exception {
+    if (nodeSet.contains(key)) {
+      containsKey(root, key);
+    } else {
+      throw new Exception("Key is out of tree");
```
brzozan: `IllegalArgumentException` would be better here and would not require `throws Exception` in method signature.

stelaval/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,176 @@
...
+   * 
+   * @return node that is the lowest common ancestor of node1 and node2
+   * 
+   * @throws Exception if either one or both nodes are not in the binary tree
+   */
+  public Node<T> getLowestCA(Node<T> node1, Node<T> node2) throws Exception {
+    if (nodeSet.contains(node1) && nodeSet.contains(node2)) {
+      return getCA(root, node1, node2);
+    } else {
+      throw new Exception("Nodes are not in the tree");
```
brzozan: `IllegalArgumentException` would be better.

stelaval/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,176 @@
...
+  public void printAncestors(Node<T> key) throws Exception {
+    if (nodeSet.contains(key)) {
+      containsKey(root, key);
+    } else {
+      throw new Exception("Key is out of tree");
+    }
+  }
+
+  /**
+   * Checks if a node is present in a subtree
```
brzozan: The description (and method name) is not exactly correct - checking if node is present in subtree is just part of what it does. More importantly it prints things out.

It would be a bit more clear if method name was `printAncestorsIfNodeInSubtree` (or `printAncestors`, or something similar) and comment contained more detailed description and the `@returns` line that would explain what is the return value.

stelaval/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,176 @@
...
+   */
+  public Node<T> getLowestCA(Node<T> node1, Node<T> node2) throws Exception {
+    if (nodeSet.contains(node1) && nodeSet.contains(node2)) {
+      return getCA(root, node1, node2);
+    } else {
+      throw new Exception("Nodes are not in the tree");
+    }
+  }
+
+  private Node<T> getCA(Node<T> currentNode, Node<T> node1, Node<T> node2) {
```
brzozan: `getCA` suggests that it is any common ancestor, not necessarily lowest. This method and the `getLowestCA` above have different parameters, so it would be safe to call this one `getLowestCA` as well, which would make it clear that it is still the lowest CA that we are looking for.

stelaval/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,176 @@
...
+      Node<Integer> root = new Node<>(0);
+      Node<Integer> node1 = new Node<>(1);
+      Node<Integer> node2 = new Node<>(2);
+      Node<Integer> node3 = new Node<>(3);
+      Node<Integer> node4 = new Node<>(4);
+      Node<Integer> node5 = new Node<>(5);
+      Node<Integer> node6 = new Node<>(6);
+      Node<Integer> node7 = new Node<>(5);
+      BinaryTree<Integer> bTree = new BinaryTree<>();
+      root.setLeft(node1);
```
brzozan: This manual tree construction is not really convenient to use. It would be much better if `Node` had methods (or one method) to add children with given values.
One of possible solutions:
```
class Node<T> {
  /** Adds node with given value as left child and returns newly created child Node object. */
  public Node<T> addLeftChild(T value) { ... }
  ...
}
```
This would let you build tree as follows:
```
Node<Integer> root = new Node<>(0);
Node<Integer> node1 = root.addLeftChild(1);
Node<Integer> node2 = root.addRightChild(2);
...
```
It would also let you get rid of `setLeft` and `setRight` methods which make it possible for someone to break your tree structure.

Even more convenient solution: have whole creating node and adding it to the tree logic inside `BinaryTree` class, so that all that user has to do to create node and add it to the tree is calling a single method rather than two.

stelaval/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,176 @@
...
+      return currentNode;
+    } else if (leftCA != null) {
+      return leftCA;
+    } else {
+      return rightCA;
+    }
+  }
+
+
+  private static class Node<T> {
```
brzozan: The whole class interface currently relies on this class being available. If you make it private, your class is impossible to use for anyone else.

stelaval/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,176 @@
...
+
+  /**
+   * Prints all the ancestors of a node
+   * 
+   * @param key is a node which ancestors are printed
+   * 
+   * @throws Exception if given key is not present in a tree
+   */
+  public void printAncestors(Node<T> key) throws Exception {
+    if (nodeSet.contains(key)) {
```
brzozan: What conditions would have to be met to avoid having such checks? It is good to do sanity checks like that in code, but it is even better to design class in a way that would make them impossible to happen.

stelaval/assignment2/BinaryTree.java
```diff
@@ -0,0 +1,176 @@
+package assignment2;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class BinaryTree<T> {
+
+  private Node<T> root;
+  private Set<Node<T>> nodeSet;
```
brzozan: It does not change, so it's better to make it `final`.

brzozan (changes requested):
Very good PR, thank you!

Algorithmically it looks OK, but I had a few comments about code structure and interface, which can be summarized as "always assume that someone else will use your code in their project, so please make it as easy and convenient to use as possible".

brzozan (approved):
Thank you!

PR #15, opened 2018-06-18
Solution to assignment 3

YakimS:


YakimS/assignment3/Dictionary.java
```diff
@@ -0,0 +1,35 @@
+package com.shaya;
+
+import java.util.Set;
+
+public class Dictionary
+{
+    private Set<String> prefixes, words;
```
brzozan: Please declare each field separately. Such merged declarations are hard to read.

YakimS/assignment3/Main.java
```diff
@@ -0,0 +1,6 @@
+package com.shaya;
+
+public class Main {
+    public static void main(String[] args) {
```
brzozan: This class does not do anything, so it seems unnecessary. Could you remove it, please? Or make it do something useful.

YakimS/assignment3/WordSearcher.java
```diff
@@ -0,0 +1,104 @@
...
+     * @param dictionary - a dictionary which contains the word and the prefixes of the words
+     * @param charsGrid -  grid of chars.
+     * @param x - marks the current X position in the grid
+     * @param y - marks the current Y position in the grid
+     * @param prefix - current prefix. The method will find words starting with this string.
+     * @param foundWords - set of all the words that are already found in the grid.
+     */
+    private void allWordsInGrid(Dictionary dictionary, char[][] charsGrid, int x, int y, String prefix, HashSet<String> foundWords)
+    {
+        if(     x < 0 || x >= charsGrid.length || y < 0 || y >= charsGrid[0].length  // if out of bound of the grid
```
brzozan: This is very odd formatting. Could you please use consistent formatting everywhere? It will make code a bit easier to read.

YakimS/assignment3/WordSearcher.java
```diff
@@ -0,0 +1,104 @@
...
+                || !dictionary.isPrefix(prefix)                                      // if this path not lead to a word
+                || charsGrid[x][y] == '0')                                           // if char in this location has been used in the current prefix
```
brzozan: This check does not work. You use different value as marker that field was used.

YakimS/assignment3/WordSearcher.java
```diff
@@ -0,0 +1,104 @@
...
+        // Create new prefix by adding current char
+        String currentPrefix = prefix + charsGrid[x][y];
+
+        // Create updated grid which that indicate that the current char
+        // Had been used in the currentPrefix
+        char[][] currentCharGrid = cloneArray(charsGrid);
+        currentCharGrid[x][y] = 0;
+
+        // Search for word with the updated prefix and grid horizontally/vertically/diagonally
+        allWordsInGrid(dictionary,currentCharGrid ,x +1 ,y +1, currentPrefix, foundWords);
```
brzozan: Could would be more maintainable if you used nested loops instead of copying method calls. Currently person that reads the code needs to mentally keep track of all the offsets added/subtracted from x and y to figure out what is going on.

YakimS/assignment3/WordSearcher.java
```diff
@@ -0,0 +1,104 @@
...
+
+
+    /**
+     * Clone 2-dim array of chars
+     *
+     * @param src - the array which will be copied
+     * @return a clone of the given array
+     */
+    private char[][] cloneArray(char[][] src) {
+        int length = src.length;
```
brzozan: Not needed. Just use `src.length` when needed. It will make it clear that length is always equal to `src.length` and compiler will optimize it away, so it will not be any slower.

YakimS/assignment3/WordSearcherTest.java
```diff
@@ -0,0 +1,36 @@
...
+    @org.junit.Test
+    public void allWordsInGrid()
+    {
+        char[][] charsGrid = new char[][]
+                {
+                        {'A', 'A', 'R'},
+                        {'T', 'C', 'D'},
+                };
+
+        HashSet wordsSet = new HashSet();
```
brzozan: Please use generic parameters to make sure that your code is type safe.

YakimS/assignment3/WordSearcherTest.java
```diff
@@ -0,0 +1,36 @@
...
+        prefixesSet.add("CART");
+        prefixesSet.add("CAT");
+        prefixesSet.add("CA");
+        prefixesSet.add("C");
+        prefixesSet.add("");
+        Dictionary dictionary = new Dictionary(prefixesSet,wordsSet);
+
+        WordSearcher wordSearcher = new WordSearcher();
+
+        System.out.print(wordSearcher.allWordsInGrid(dictionary,charsGrid));
```
brzozan: Tests should verify correctness of results automatically. Typically it is done using `assert` family of methods provided by JUnit's [`Assert` class](https://junit.org/junit4/javadoc/4.12/org/junit/Assert.html).
In this particular situation the best match would probably be combining `Assert.assertThat` and [`IsCollectionContaining.hasItems`](https://junit.org/junit4/javadoc/4.12/org/hamcrest/core/IsCollectionContaining.html#hasItems(T...))

YakimS/assignment3/WordSearcherTest.java
```diff
@@ -0,0 +1,36 @@
+package com.shaya;
+
+import java.util.HashSet;
+
+
+public class WordSearcherTest {
+
+    @org.junit.Test
```
brzozan: Just import `org.junit.Test` and use `@Test` as annotations.

YakimS/assignment3/WordSearcherTest.java
```diff
@@ -0,0 +1,36 @@
+package com.shaya;
+
+import java.util.HashSet;
+
+
+public class WordSearcherTest {
```
brzozan: It would be good to include tests for some corner cases...

brzozan (changes requested):
Thank you for pull request. I added a few comments in code.

YakimS/assignment3/WordSearcher.java
YakimS: I dont see the second maker. Did i use a different marker than the char '0'?

YakimS:
Thank you very much for the review!
This is the first time I ever used JUnit, so your feedbacks are even extra helpful.
I have done the requested changes, except the one regarding the marking. I have commented on your note.

brzozan (approved):
LGTM, thank you!

PR #16, opened 2018-06-18
narciuch 3

narciuch:


narciuch/assignment3/narciuch-3/src/WordSearch.java
```diff
@@ -0,0 +1,145 @@
...
+   * <li>We cannot visit the same cell more than once in the same word.</li>
+   * </ul>
+   *
+   * @param grid of letters in which to look for words from the dictionary
+   * @param dict the dictionary of words to look for
+   * @return list of words from the dictionary that can be formed in the grid
+   */
+  public List<String> findWords(char[][] grid, Dictionary dict) {
+    List<String> wordsFound = new ArrayList<>();
+    if (grid.length == 0) {
```
brzozan: Good handling of a corner case!

Optional: handle the other corner case (`grid[0].length`) here explicitly as well rather than relying on loops not to process any elements. This would make code behaviour a bit more clear.

narciuch/assignment3/narciuch-3/src/WordSearch.java
```diff
@@ -0,0 +1,145 @@
...
+   * @param letters    the letters we have visited on our current path through the grid
+   * @param visited    an array with the same dimensions as the grid.
+   *                   The boolean entry at row i, column j corresponds to whether we have
+   *                   already visited the letter at row i, column j in the grid, and hence
+   *                   should not revisit it.
+   * @param row        the row of the cell currently being considered as the start point of the possible words
+   * @param column     the column of the cell currently being considered as the start point of the possible words
+   * @param wordsFound list of words we have already been able to form in the grid
+   */
+  private void findFurtherWords(char[][] grid, Dictionary dict, List<Character> letters,
```
brzozan: Code would be simpler if you used `String` instead of `List<Character>` for `letters`. It would remove need to convert list to string and need to remove last element at the end of this method.

narciuch/assignment3/narciuch-3/src/WordSearch.java
```diff
@@ -0,0 +1,145 @@
...
+    int nr_columns = grid[0].length;
+    visited[row][column] = true;
+    letters.add(grid[row][column]);
+
+    //For the case of one letter words
+    if (dict.isWord(charListToString(letters))) {
+      wordsFound.add(charListToString(letters));
+    }
+
+    int[][] neighbours = {{row - 1, column - 1}, {row - 1, column}, {row - 1, column + 1},
```
brzozan: It would be much cleaner if you user nested `for` loops to iterate over neighbours - currently person reading code has to mentally track 8 cases to make sure that offsets are correct.

narciuch/assignment3/narciuch-3/test/DictionaryTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+import static junit.framework.TestCase.assertTrue;
+
+public class DictionaryTest {
+  String[] words = {"CAR", "CARD", "CART", "CAT"};
+  Dictionary dict = new Dictionary(words);
+
+  @Test
+  public void correctlyIdentifiesWord() {
+    assert(dict.isWord("CAR"));
```
brzozan, line 11: I think you meant `assertTrue`, since you already imported it?

`assert` kind of works, but it is better for sanity checks in main code rather than tests.

narciuch/assignment3/narciuch-3/test/WordSearchTest.java
```diff
@@ -0,0 +1,66 @@
...
+public class WordSearchTest {
+  String[] words = {"CAR", "CARD", "CART", "CAT"};
+  Dictionary dict = new Dictionary(words);
+  WordSearch wordSearch = new WordSearch();
+  char[][] grid = {{'A', 'A', 'R'}, {'T', 'C', 'D'}};
+
+  @Test
+  public void findsWordsInGrid() {
+    List<String> wordsFound = wordSearch.findWords(grid, dict);
+    assert(wordsFound.contains("CAR"));
```
brzozan: `assertTrue`.

Optional: Also please take a look at the `assert` family of methods in [`Assert` class](https://junit.org/junit4/javadoc/4.12/org/junit/Assert.html), especially `Assert.assertThat` and [`IsCollectionContaining.hasItems`](https://junit.org/junit4/javadoc/4.12/org/hamcrest/core/IsCollectionContaining.html#hasItems(T...))

If you use static imports you can get nice assertions that read almost like a sentence.

brzozan (changes requested):
Thank you for a very good PR! I added some relatively minor comments in code.

narciuch/assignment3/narciuch-3/src/WordSearch.java
narciuch: Will do for clarity

narciuch/assignment3/narciuch-3/src/WordSearch.java
narciuch: Agreed. I initially thought it would be more efficient to use a list since we are changing it a lot, but have since realised immutability should be ensured.

narciuch/assignment3/narciuch-3/test/DictionaryTest.java
narciuch: Interesting observation. Will do

narciuch/assignment3/narciuch-3/src/WordSearch.java
narciuch: Yes, will do. Was not sure how best to represent and iterate through all valid neighbour cells.

brzozan (approved):
LGTM. Thank you!

PR #17, opened 2018-06-19
Assignment 3

VolhaAnishchanka:


HelgaShiryaeva/assignment3/DictionaryTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Set;
+import java.util.HashSet;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class DictionaryTest {
+    private Dictionary dictionary;
+    private WordSearch wordSearch;
```
brzozan: `WordSearch` should have its own test class. Idea behind unit tests is that they test a single unit of code, which in case of Java is typically a single class.

HelgaShiryaeva/assignment3/DictionaryTest.java
```diff
@@ -0,0 +1,61 @@
...
+    @Test
+    public void isPrefixTest() {
+        assertTrue(dictionary.isPrefix("c"));
+        assertTrue(dictionary.isPrefix("ca"));
+        assertTrue(dictionary.isPrefix("cat"));
+        assertFalse(dictionary.isPrefix("fdkl"));
+    }
+
+    @Test
+    public void wordSearchTest() {
```
brzozan: It would be good to verify edge cases too.

HelgaShiryaeva/assignment3/DictionaryTest.java
```diff
@@ -0,0 +1,61 @@
...
+        char[][] grid = new char[2][3];
+        grid[0][0] = 'a'; grid[0][1] = 'a'; grid[0][2] = 'r';
```
brzozan: Simpler version:
```
char[][] grid = new char[][] {
  {'a', 'a', 'r'},
  {'t', 'c', 'd'}
};
```

HelgaShiryaeva/assignment3/WordSearch.java
```diff
@@ -0,0 +1,90 @@
...
+        prefix.append(board[i][j]);
+        if(!dictionary.isPrefix(prefix.toString())) {
+            prefix.deleteCharAt(prefix.length() - 1);
+            return;
+        }
+        if(dictionary.isWord(prefix.toString())) {
+            words.add(prefix.toString());
+        }
+        isVisited[i][j] = true;
+        dfs(board, i, j - 1, prefix, words, isVisited);
```
brzozan: Code would be easier to understand if you used nested loops to iterate over offsets. With current list of almost-identical-but-not-really calls to dfs person reading code needs to mentally keep track of 8 sets of param values to make sure that correct fields are visited and it is easy to miss a bug if at one place there is a '+' instead of '-'.

brzozan (changes requested):
Very, very clean and well commented PR. Thank you!

I added some relatively minor comments.

brzozan (approved):
LGTM, thank you!

VolhaAnishchanka:
Thank you so much for the review ! :)

PR #18, opened 2018-06-19
Assignment 3

stelaval:


Dictionary.java
```diff
@@ -0,0 +1,28 @@
+package assignment3;
+
+import java.util.Set;
+
+public class Dictionary {
+  private Set<String> words;
```
brzozan: `final`

WordSearch.java
```diff
@@ -0,0 +1,80 @@
...
+
+    int columns = grid[0].length;
+    if (columns == 0) {
+      return wordsFound;
+    }
+
+    for (int r = 0; r < rows; r++) {
+      for (int c = 0; c < columns; c++) {
+        boolean[][] isVisited = new boolean[rows][columns];
+        String currentString = "";
```
brzozan: No need to have it in separate variable. You can just pass `""` as argument.

WordSearch.java
```diff
@@ -0,0 +1,80 @@
...
+    int columns = grid[0].length;
+    isVisited[row][col] = true;
+    currentString += grid[row][col];
+
+    if (dictionary.isWord(currentString)) {
+      wordsFound.add(currentString);
+    }
+
+    if (dictionary.isPrefix(currentString)) {
+      for (int r = row - 1; r <= row + 1 && r < rows; r++) {
```
brzozan: It would be a bit cleaner if the `r < rows` and `c < columns` were checked in the `if` statement in line 71.

I know that it is a small optimization to avoid iterating over columns if row is outside of the grid, but it feels quite inconsistent that half of the condition is checked here and half in line 71.

WordSearchTest.java
```diff
@@ -0,0 +1,54 @@
...
+    HashSet<String> words = new HashSet<>();
+    words.add("CAR");
+    words.add("CARD");
+    words.add("CART");
+    words.add("CAT");
+    Dictionary dictionary = new Dictionary(words);
+
+    Set<String> wordsFound = WordSearch.findAllWords(grid, dictionary);
+
+    assert (wordsFound.contains("CAR"));
```
brzozan: Typically you would use one of the methods from [`org.junit.Assert`](http://junit.sourceforge.net/javadoc/org/junit/Assert.html) class.

While `assert` keyword kind of works (i.e. fails), methods from `Assert` are extremely useful and can provide additional information about what exactly went wrong.

WordSearchTest.java
```diff
@@ -0,0 +1,54 @@
...
+    HashSet<String> words = new HashSet<>();
+    words.add("CAR");
+    words.add("CARD");
+    words.add("CART");
+    words.add("CAT");
+    Dictionary dictionary = new Dictionary(words);
+
+    Set<String> wordsFound = WordSearch.findAllWords(grid, dictionary);
+
+    assert (wordsFound.size() == 0);
```
brzozan: Even better: `assertTrue(wordsFound.isEmpty());`. It's a minor difference, but immediately tells person who reads the code that container is expected to be empty, without the need to think about what does size equal 0 mean.

WordSearchTest.java
```diff
@@ -0,0 +1,54 @@
...
+    words.add("CART");
+    words.add("CAT");
+    Dictionary dictionary = new Dictionary(words);
+
+    Set<String> wordsFound = WordSearch.findAllWords(grid, dictionary);
+
+    assert (wordsFound.contains("CAR"));
+    assert (wordsFound.contains("CAT"));
+    assert (wordsFound.contains("CARD"));
+    assert (!wordsFound.contains("CAP"));
```
brzozan: Good! It is important to not only check that expected words were returned, but also the opposite - that no unexpected words were there.

Stronger versions of this check:

- check for positives (expected words) and then also verify size of the container to make sure that it does not include any more words.
- create a set of expected words and use `assertEquals(wordsExpected, wordsFound)`.

brzozan (changes requested):
Very good and clean PR! Thank you!

I added a few comments, but they are rather minor and focus on the testing side.

brzozan (approved):
LGTM, thank you!

PR #19, opened 2018-06-20
Submit Assignment 3

yujo-t:


uro98/assignment3/Dictionary.java
```diff
@@ -0,0 +1,59 @@
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Represents a collection of words.
+ */
+public class Dictionary {
+  private Set<String> wordSet = new HashSet<String>();
```
brzozan: `final`, since those fields do not change over lifetime of the object.

uro98/assignment3/Dictionary.java
```diff
@@ -0,0 +1,59 @@
...
+  private Set<String> prefixSet = new HashSet<String>();
+
+  public Set<String> getWordSet() {
+    return wordSet;
+  }
+  
+  public Set<String> getPrefixSet() {
```
brzozan: `getPrefixSet` seems to not be used anywhere, so it is better to remove it.

uro98/assignment3/FindAllWords.java
```diff
@@ -0,0 +1,80 @@
...
+            i,
+            j);
+        totalWords.addAll(wordSet);
+      }
+    }
+    
+    return totalWords;
+  }
+  
+  public static Set<String> searchWords(
```
brzozan: I think this should be private, since it is an implementation detail. From user's perspective `findAllWords` is the only interface they need.

Also: even though it should be a private method, its interface is quite complicated, so it would be good to add comment to explain how it should be used.

uro98/assignment3/FindAllWords.java
```diff
@@ -0,0 +1,80 @@
...
+      Grid grid,
+      Dictionary dictionary,
+      boolean[][] visited,
+      Set<String> words,
+      List<String> path,
```
brzozan: It is not necessary to store `String`s as path elements. `Character` would be enough.

Even better - just use `String` rather than `List<String>` - this will simplify code quite a bit since you will not need to join elements and make explicit copies.

uro98/assignment3/FindAllWords.java
```diff
@@ -0,0 +1,80 @@
...
+    for (int i = -1; i <= 1; i++) {
+      for (int j = -1; j <= 1; j++) {
+        // Make copies of path and visited to search for words in different states
+        List<String> pathCopy = new ArrayList<String>(path);
+        boolean[][] visitedCopy = new boolean[visited.length][];
+        for (int k = 0; k < visited.length; k++) {
+          visitedCopy[k] = visited[k].clone();
+        }
+        
+        Set<String> newWords = searchWords(grid, dictionary, visitedCopy, words, pathCopy, posX + i, posY + j);
```
brzozan: With this invocation, `newWords` does not actually contain new words. Depending on situation it contains one of:

- empty set if no new words have been found;
- set of all words found so far (both old and new), since you are passing the same `Set` object created in line 24 all the way down and modify it on the way.

There are a few options:

- Don't pass `words` as parameter at all and just create empty `Set` at the beginning of this method.
- Don't return final set from this method at all and just rely on the fact, that code will modify set that you passed here from line 24.
- Keep returning final set as you do now, since it is convenient in the place where you start recursion, but then you do not need `newWords` variable and no need to call `words.addAll(newWords)` below.

I think the first option is cleanest, but all of them are valid, so it is up to you which one to choose.

uro98/assignment3/Grid.java
```diff
@@ -0,0 +1,58 @@
+/**
+ * Represents a grid of letters.
+ */
+public class Grid {
+  private int height;
+  private int width;
+  private String[][] grid;
```
brzozan: `char` would be better - way more efficient.

Also: fields could be `final`.

uro98/assignment3/FindAllWordsTest.java
```diff
@@ -0,0 +1,66 @@
+import static org.junit.Assert.*;
```
brzozan: Please avoid importing via asterisk.

uro98/assignment3/FindAllWordsTest.java
```diff
@@ -0,0 +1,66 @@
...
+    expected.add("CAR");
+    expected.add("CARD");
+    expected.add("CAT");
+    
+    assertEquals(expected, wordSet);
+    assertFalse(dictionary.getWordSet().equals(wordSet));
+  }
+
+  @Test
+  public void testSearchWords() {
```
brzozan: Not really needed. `searchWords` is an implementation detail and should be private, and tests typically use only public interface. There are rare cases when tests need to use normally hidden features of the class, but they are very rare.

uro98/assignment3/FindAllWordsTest.java
```diff
@@ -0,0 +1,66 @@
...
+    
+    Set<String> wordSet = FindAllWords.findAllWords(grid, dictionary);
+    
+    Set<String> expected = new HashSet<String>();
+    expected.add("CAR");
+    expected.add("CARD");
+    expected.add("CAT");
+    
+    assertEquals(expected, wordSet);
+    assertFalse(dictionary.getWordSet().equals(wordSet));
```
brzozan: Not necessary. If you want to test `Dictionary` (good!), it should have separate class with tests, so that one test class focuses on only one tested class.

uro98/assignment3/FindAllWordsTest.java
```diff
@@ -0,0 +1,66 @@
+import static org.junit.Assert.*;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.junit.Test;
+
+public class FindAllWordsTest {
```
brzozan, line 9: It would be great to add tests for edge cases too. What would be some other good scenarios to check to make sure that findWords returns only words that exist in the dictionary and only ones that can actually be constructed from the grid?

brzozan (changes requested):
Very nice PR. Well done!

I added a few comments in code.

uro98/assignment3/FindAllWords.java
yujo-t: Thank you for the detailed feedback. Just out of curiosity, why do you think the first option is the cleanest?

uro98/assignment3/FindAllWordsTest.java
yujo-t: Are private methods usually not unit tested then? Why shouldn't they be?

uro98/assignment3/FindAllWordsTest.java
yujo-t: Do you change your imports after you have finished coding or you don't use * at all? Importing * in Eclipse would show recommendations while coding which I think is useful, but I can imagine it'd be hard to remember to change imports afterwards.

uro98/assignment3/FindAllWordsTest.java
yujo-t: Apart from empty grids and dictionaries, doesn't the first test cover both of those scenarios?

brzozan: You are right. Checking whether returned set of words is equal to set of expected words does indeed verify that.

uro98/assignment3/FindAllWordsTest.java
brzozan: Private methods are usually not tested directly, only via public interface of the class. Reasoning: if public methods work correctly, then private methods that are used to implement them, work too. Of course then one has to make sure that test data set is sufficient to trigger all interesting paths in private methods.

Of course you can also do the trick that was done in Assignment 3 presentation (Laura should have shared it with you in "solutions" channel on Slack) - make those methods not private and not public, but package private. This lets you test them as long as test class is in the same package. Downside is that this breaks hermetization and exposes things that should normally be implementation detail.

There are also more advanced tricks like `@VisibleForTesting` annotation, which makes method visible in tests, but in other code it will be visible as private. Or approach that "if a method is complex enough to have dedicated set of tests, then maybe it should be part of separate class, where it will be public".

In general there is a balance on keeping interface of the class small and focused, and keeping it easily testable.

If you want to keep your tests for helper methods, feel free to keep them, just make the methods package private.

brzozan (changes requested):
Thank you for the changes! LGTM.

uro98/assignment3/FindAllWordsTest.java
yujo-t: I see, thank you for the detailed answer!

brzozan (approved):

PR #21, opened 2018-06-29
Upload Assignment4 files

YakimS:


YakimS/assignment4/IslandsCounter.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Arrays;
+
+public class IslandsCounter {
+    /**
+     *  Counts the number of islands in a 2-dimensional array,
+     *  which each tile of it is either land(=true) or water(=false).
+     *  Two tiles belong to the same island if they are both land and are
+     *  adjacent horizontally or vertically, but not diagonally.
+     */
+    public int countIsland(boolean[][] boolMap) {
```
brzozan: Better name for the parameter would be just `map`.

Also nitpick about method name: `countIslands` would be better.

YakimS/assignment4/IslandsCounter.java
```diff
@@ -0,0 +1,55 @@
+package com.shaya.CodeUAss4Package;
+
+import java.util.Arrays;
+
+public class IslandsCounter {
+    /**
+     *  Counts the number of islands in a 2-dimensional array,
+     *  which each tile of it is either land(=true) or water(=false).
```
brzozan: Typically details about contents of parameters is done  as `@param` line rather than in the main method description. This makes it possible for IDE to parse comment and show description of parameter as a tooltip when you hover over it in code, which is really convenient.

YakimS/assignment4/IslandsCounter.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class IslandsCounter {
+    /**
+     *  Counts the number of islands in a 2-dimensional array,
+     *  which each tile of it is either land(=true) or water(=false).
+     *  Two tiles belong to the same island if they are both land and are
+     *  adjacent horizontally or vertically, but not diagonally.
+     */
+    public int countIsland(boolean[][] boolMap) {
+        int[][] islandNumMap = new int[boolMap.length][boolMap[0].length];
```
brzozan: What if somebody passes `{}`?

YakimS/assignment4/IslandsCounterTest.java
```diff
@@ -0,0 +1,88 @@
+package com.shaya.CodeUAss4Package;
+
+import org.junit.Assert;
+
+import static org.junit.Assert.*;
+
+public class IslandsCounterTest {
+
+    @org.junit.Test
```
brzozan: `import org.junit.Test;` and then `@Test`.

YakimS/assignment4/IslandsCounterTest.java
```diff
@@ -0,0 +1,88 @@
+package com.shaya.CodeUAss4Package;
+
+import org.junit.Assert;
+
+import static org.junit.Assert.*;
```
brzozan: Please do not import with asterisk...

YakimS/assignment4/IslandsCounterTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+        boolean[][] boolGrid = new boolean[][]
+                {
+                        { false , true, false, true},
+                        { true , true, false, false},
+                        { false , false, true, false},
+                        { false , false, true, false},
+                };
+
+        Assert.assertEquals(islandsCounter.countIsland(boolGrid), 3);
```
brzozan: `assertEquals` takes expected value as first parameter, not the second one. This might seem to be not important, since assertion will fail in case of error no matter what order parameters are in, but will make error message confusing.

Also: no need for `Assert.` part - all methods are already imported.

YakimS/assignment4/IslandsCounter.java
```diff
@@ -0,0 +1,55 @@
...
+
+        // If this part of the boolean map have been searched
+        if( islandNumMap[i][j] != -1 ) {
+            return;
+        }
+
+        if( !boolMap[i][j]) {
+            islandNumMap[i][j] = 0; // Indicates that this part of the map is water
+        }
+        else{
```
brzozan: `else` should be in previous line. Also: space before `{`.

YakimS/assignment4/IslandsCounterTest.java
```diff
@@ -0,0 +1,88 @@
...
+public class IslandsCounterTest {
+
+    @org.junit.Test
+    public void countIslandInExampleGrid()
+    {
+        IslandsCounter islandsCounter = new IslandsCounter();
+
+        boolean[][] boolGrid = new boolean[][]
+                {
+                        { false , true, false, true},
```
brzozan: Nitpick: spaces before closing brace.

brzozan (changes requested):
Very nice and well tested PR! I had only a few comments.

YakimS/assignment4/IslandsCounter.java
YakimS: Right! I've added the relevant test and fix.

YakimS/assignment4/IslandsCounterTest.java
YakimS: Wow, nice catch!

YakimS:
I've implemented the requested changes.
Thank you very much for the code review!

YakimS/assignment4/IslandsCounter.java
```diff
@@ -3,23 +3,32 @@
...
+     *
+     * @param map - 2-dimensional array,
+     *            which each tile of it is either land(=true) or water(=false).
+     * @return number of "islands" in the map.
      */
-    public int countIsland(boolean[][] boolMap) {
-        int[][] islandNumMap = new int[boolMap.length][boolMap[0].length];
+    public int countIslands(boolean[][] map) {
+        // make sure the given array is 2-dim
+        if(map.length==0){
```
brzozan: Nit: space after `if`.

YakimS/assignment4/IslandsCounterTest.java
```diff
@@ -1,44 +1,44 @@
...
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
 
 public class IslandsCounterTest {
 
-    @org.junit.Test
+    @Test
     public void countIslandInExampleGrid()
     {
```
brzozan: Nit: this bracket should be in the same line as method name. Same for other methods below.

brzozan (approved):
LGTM, thank you!

Before merging, please fix the minor code style comments I added.

PR #22, opened 2018-07-02
Helga Shiryaeva #4

VolhaAnishchanka:


HelgaShiryaeva/assignment4/IslandsCounter
```diff
@@ -0,0 +1,88 @@
...
+ */
+public class IslandsCounter {
+
+    /**
+     * This method is made to simplify counting connected components
+     * of a  two-dimensional grid. Breadth first search explores layer by layer the whole
+     * island of booleans with true value. The rule of exploring is that each cell is
+     * adjacent to its top, left, bottom and right neighbour.
+     * @param tiles is a grid of booleans which connected components (islands)
+     * will be counted
```
brzozan: It would be good to include information about data format (that `true` indicates land) here. This would let IDE parse it out and show as a tooltip, which is very helpful, since person that wants to use this method does not have to read full description to find out what should the contents of `tiles` be.

HelgaShiryaeva/assignment4/IslandsCounter
```diff
@@ -0,0 +1,88 @@
...
+     * It checks whether the given cell has already been explored and if it doesn't
+     * than it is marked as explored
+     * @param queue is a queue used during breadth first search
+     * @param cell is a cell which needs to be checked and explored
+     * @param tiles is a given grid of booleans which connected components are being explored
+     */
+    private void exploreAdjacentCell(Queue<Cell> queue, Cell cell, boolean[][] tiles) {
+        int row = cell.getRow();
+        int column = cell.getColumn();
+        if(!(row < 0 || column < 0 || row >= tiles.length || column >= tiles[0].length) && tiles[row][column]) {
```
brzozan: Please avoid complicating logic like that. It is quite difficult to analyse. It would be much better if it was rewritten as `(row >= 0 && column >= 0 && row < tiles.length && column < tiles.length && tiles[row][column])`.
Alternatively it could be split into two ifs - on to check coordinates, one to check if tile is `true`.

HelgaShiryaeva/assignment4/IslandsCounter
```diff
@@ -0,0 +1,88 @@
...
+        if (tiles.length == 0 || tiles[0].length == 0) {
+            throw new IllegalArgumentException("Unexpectable map of tiles");
+        }
+        int count = 0;
+        Queue<Cell> queue = new LinkedList<>();
+        for (int i = 0; i < tiles.length; i++) {
+            for (int j = 0; j < tiles[0].length; j++) {
+                if (tiles[i][j]) {
+                    queue.add(new Cell(i, j));
+                    tiles[i][j] = false;
```
brzozan, line 39: This is very dangerous and should be explicitly mentioned in method comments. User of this method might not expect that the data will get modified. Imagine if someone wanted to print out map if it included more than 3 islands. They would probably write code like:

```
booelan[][] map = getMyAwesomeMap();
if (islandCounter.countIslands(map) > 3) {
  printMap(map); // what will be printed?
}
```

HelgaShiryaeva/assignment4/IslandsCounter
```diff
@@ -0,0 +1,88 @@
...
+            queue.add(new Cell(row, column));
+            tiles[row][column] = false;
+        }
+    }
+
+    /**
+     * This class represents a cell of a two-dimensional grid
+     */
+    class Cell {
+        int row;
```
brzozan: Both those fields should be private and final.

HelgaShiryaeva/assignment4/IslandsCounterTest
```diff
@@ -0,0 +1,105 @@
...
+
+    @Test
+    public void threeComponentsTest() {
+        boolean tiles[][] = {
+                {false, true, false, true},
+                {true, true, false, false},
+                {false, false, true, false},
+                {false, false, true, false}
+        };
+        IslandsCounter islandsCounter = new IslandsCounter();
```
brzozan: `IslandsCounter` is stateless, so it typically would be created either directly as a field in `IslandsCounterTest` or in `setUp()` method.

brzozan (changes requested):
Very nice PR! Solution with using queue is interesting, even though a bit more difficult to analyse.

I added a few comments to the code.

brzozan (approved):
Thank you! LGTM.

PR #23, opened 2018-07-02
Implement island finder

narciuch:


narciuch/assignment4/src/IslandFinder.java
```diff
@@ -0,0 +1,132 @@
...
+   * <p>Two tiles in the matrix are part of the same island if
+   * they are both land and are adjacent horizontally or vertically.
+   *
+   * @param nrRows    number of rows in the lands matrix
+   * @param nrColumns number of columns in the lands matrix
+   * @param lands     nested boolean array, with true representing land,
+   *                  false representing water
+   * @return the number of islands in the lands matrix
+   */
+  public int getIslands(int nrRows, int nrColumns, boolean[][] lands) {
```
brzozan: nrRows and nrColumns params are not really needed - dimensions can be extracted from array itself. If you want to keep them, there should be check if those values match actual dimensions of the array.

narciuch/assignment4/test/IslandFinderTest.java
```diff
@@ -0,0 +1,33 @@
...
+import static junit.framework.TestCase.assertTrue;
+
+public class IslandFinderTest {
+  IslandFinder islandFinder = new IslandFinder();
+
+  @Test
+  public void findsMultipleIslands() {
+    boolean[][] land = {{false, true, false, true}, {true, true, false, false},
+      {false, false, true, false}, {false, false, true, false}};
+    assertTrue(islandFinder.getIslands(4, 4, land) == 3);
```
brzozan: I think it is better to use `assertEquals` from `org.junit.Assert` class. This will give you better error messages in case of test failure ("not true that bad_value is equal to expected_value" or sth similar).

narciuch/assignment4/src/IslandFinder.java
```diff
@@ -0,0 +1,132 @@
...
+   * @param islands       a list of integer lists, where each inner integer list represents an island
+   * @param islandIndexes the indices of those inner arrays in 'islands' that are to be merged
+   * @return a new copy of islands which has merged the islands as specified and not changed the
+   * other islands within the 'islands' nested list
+   */
+  private List<List<Integer>> mergeIslands(List<List<Integer>> islands, List<Integer> islandIndexes) {
+    List<Integer> mergeInto = islands.get(islandIndexes.get(0));
+    for (int mergeIslandIndex = 1; mergeIslandIndex < islandIndexes.size(); mergeIslandIndex++) {
+      int islandNr = islandIndexes.get(mergeIslandIndex);
+      for (int tileNr : islands.get(islandNr)) {
```
brzozan: Why not just `mergeInto.addAll(islands.get(islandNr);`?

narciuch/assignment4/src/IslandFinder.java
```diff
@@ -0,0 +1,132 @@
...
+   * @param nrColumns number of columns in the lands matrix
+   * @param lands     nested boolean array, with true representing land,
+   *                  false representing water
+   * @return the number of islands in the lands matrix
+   */
+  public int getIslands(int nrRows, int nrColumns, boolean[][] lands) {
+    if (lands.length == 0 || lands[0].length == 0) {
+      return 0;
+    }
+    List<List<Integer>> islands = new ArrayList<>();
```
brzozan: I feel like code could be simplified by using some sort of map, potentially having unique pos as key. Did you consider such approach?

narciuch/assignment4/test/IslandFinderTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import static junit.framework.TestCase.assertTrue;
+
+public class IslandFinderTest {
+  IslandFinder islandFinder = new IslandFinder();
+
+  @Test
+  public void findsMultipleIslands() {
+    boolean[][] land = {{false, true, false, true}, {true, true, false, false},
+      {false, false, true, false}, {false, false, true, false}};
```
brzozan: Map would be easier to read if it was formatted as one row per line with values aligned in columns.

brzozan (changes requested):
Thanks for a really nice PR! I have a few relatively minor comments about the code and one question about algorithm implementation.
Please see discussion for details.

narciuch/assignment4/src/IslandFinder.java
narciuch: Yes, I completely agree. I did this because in the specification it was written that "The input to your function is the number of rows, number of columns, and a 2-dimensional array of booleans, where false means water and true means land." I will change the code t just extract row and column numbers from array itself as this definitely makes more sense.

narciuch/assignment4/test/IslandFinderTest.java
narciuch: Agreed.

narciuch/assignment4/src/IslandFinder.java
narciuch: Yes, much simpler.

narciuch/assignment4/src/IslandFinder.java
```diff
@@ -1,26 +1,24 @@
...
-import java.util.List;
-import java.util.stream.Collectors;
+import java.util.*;
 
 public class IslandFinder {
 
   /**
    * Returns the number of islands in the lands matrix.
-   *
+   * <p>
```
brzozan: Not really needed. There is a paragraph already in line 8.

narciuch/assignment4/src/IslandFinder.java
```diff
@@ -69,43 +67,40 @@ public int getIslands(int nrRows, int nrColumns, boolean[][] lands) {
...
+   * @return key of the island containing the tilePosition value if it exists in 'islands'.
+   * If the tilePosition is not in 'islands' then -1 is returned.
    */
-  private int getIslandIndex(List<List<Integer>> islands, int tilePosition) {
-    int totalIslandNr = islands.size();
-    for (int islandNr = 0; islandNr < totalIslandNr; islandNr++) {
-      if (islands.get(islandNr).contains(tilePosition)) {
-        return islandNr;
+  private int getIslandKey(Map<Integer, Set<Integer>> islands, int tilePosition) {
+    for (int key : islands.keySet()) {
```
brzozan: Optional: if you used `Map.entrySet()` for iteration, you could avoid looking key up with `get(key)` below.

brzozan (approved):
LGTM, thank you!

PR #24, opened 2018-07-03
Submit Assignment 4

yujo-t:


uro98/assignment4/CountIslands.java
```diff
@@ -0,0 +1,57 @@
...
+public class CountIslands {
+
+  /**
+   * Returns the number of islands in a given map.
+   * @param rowCount  the number of rows in the map
+   * @param colCount  the number of columns in the map
+   * @param map       the boolean map where true is land and false is water
+   * @return the number of islands in integer
+   */
+  public static int countIslands(int rowCount, int colCount, boolean[][] map) {
```
brzozan: rowCount and colCount are not really needed - dimensions can be extracted from the map itself. If you want to keep them, it would be good to check if values passed as rowCount and colCount match actual dimensions of the map to avoid bugs.

uro98/assignment4/CountIslands.java
```diff
@@ -0,0 +1,57 @@
...
+
+  private static boolean findIsland(int y, int x, int rowCount, int colCount, boolean[][] map) {
+
+    if (isOutOfBounds(y, x, rowCount, colCount)) {
+      return false;
+    }
+
+    if (map[y][x] == true) {
+      // Disable the land as it has been found
+      map[y][x] = false;
```
brzozan, line 60: This is very dangerous and should be mentioned in comments of the `countIslands` method. If someone uses `countIslands` in their code they might not expect that their data will get modified.
Imagine that I want to print maps that contain at least 3 islands. I could write code like this:

```
boolean[][] map = getMyAwesomeMap(3, 4);
if (CountIslands.countIslands(3, 4, map) >= 3) {
  printMap(map); // what will be printed here?
}
```

uro98/assignment4/CountIslands.java
```diff
@@ -0,0 +1,57 @@
...
+      
+      // The surrounding land is part of the same island
+      disableSurroundingLand(y, x, rowCount, colCount, map);
```
brzozan: Name is a bit confusing. Something like `visitSurroundingTiles` or something similar would be a bit better.

brzozan (changes requested):
Very nice PR! I added a few comments.

uro98/assignment4/CountIslands.java
yujo-t: I'd like to keep them because the assignment doc specified that they are part of the function input. I'll add dimension checks.

brzozan (approved):
LGTM, thank you!

PR #25, opened 2018-07-03
Assignment 4

stelaval:


stelaval/assignment4/IslandsCounter.java
```diff
@@ -0,0 +1,57 @@
...
+      throw new IllegalArgumentException("Map of tiles is empty.");
+    }
+    int rows = map.length;
+    int columns = map[0].length;
+    boolean[][] visited = new boolean[rows][columns];
+
+    int count = 0;
+    for (int i = 0; i < rows; i++) {
+      for (int j = 0; j < columns; j++) {
+        if (map[i][j] == true && !visited[i][j]) {
```
brzozan, line 27: The `== true` part is not really necessary. Feel free to leave it though if you find it more readable.

stelaval/assignment4/IslandsCounterTest.java
```diff
@@ -0,0 +1,64 @@
+package assignment4;
+
+import static org.junit.Assert.*;
```
brzozan: Please do not import with asterisk...

stelaval/assignment4/IslandsCounterTest.java
```diff
@@ -0,0 +1,64 @@
...
+    IslandsCounter counter = new IslandsCounter();
+    
+    boolean[][] map = new boolean[][] {
+      {false, true, false, true},
+      {true, true, false, false},
+      {false, false, true, false},
+      {false, false, true, false}
+    };
+    
+    assertEquals (counter.countIslands(map), 3);
```
brzozan: `assertEquals` assumes that expected value is passed as the first parameter and the actual value as the second. If you pass them the other way round, tests will still fail in case of error, but the error message will be confusing since it will mention incorrect value as expected one.

stelaval/assignment4/IslandsCounterTest.java
```diff
@@ -0,0 +1,64 @@
+package assignment4;
+
+import static org.junit.Assert.*;
+import org.junit.Test;
+
+public class IslandsCounterTest {
+  
+  @Test
+  public void countIslandsExample() {
+    IslandsCounter counter = new IslandsCounter();
```
brzozan: IslandsCounter is stateless, so it can safely be created either in `setUp()` method or directly as field in `IslandsCounterTest` class.

brzozan (approved):
Very good PR. I had a few comments on the testing code, but other than that LGTM.

Please fix the tests and after that feel free to merge it. Good job!

PR #26, opened 2018-07-19
Group Solution to assignment #5

VolhaAnishchanka:
Some comments on the code still need to be added


group/assignment5/AlphabetDiscoverer.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+        graph = new Graph<>();
+        addAllCharsInDictionaryToGraph(dictionary);
+        addConstraints(dictionary);
+    }
+
+    private void addAllCharsInDictionaryToGraph(List<String> dictionary) {
+        for (String s : dictionary) {
+            for (int i = 0 ; i < s.length() ; i++) {
+                Character vertex = s.charAt(i);
```
YakimS: The character should be called "currentChar" or similar name, but not "vertex".
This is because this Character doesn't represent a vertex, but a value which a vertex will hold.

group/assignment5/AlphabetDiscoverer.java
```diff
@@ -0,0 +1,86 @@
...
+                if (!graph.containsVertex(vertex)) {
```
YakimS: I think we should move this check to the function "addVertex" in Graph class, and comment in the class comment that says: "Graph vertices are unique - only one vertex can be created for each T value".

group/assignment5/Graph.java
```diff
@@ -0,0 +1,218 @@
...
+        return vertecies.containsKey(value);
+    }
+
+    /**
+     * This method is to find one of possible topological
+     * orders (if a graph has more than one) of a graph vertecies
+     * @return topological ordered list of vertecies of a graph
+     */
+    public List<T> getTopologicalOrder() {
+        List<T> alphabet = new ArrayList<>();
```
YakimS: Shouldn't be called "alphabet" because the List<T> can represent many things and not just alphabets.
A proper name, in my opinion, is "verticesValues".
The use of the word "alphabet" should be changed in all the functions in Graph.class

group/assignment5/Constraint.java
```diff
@@ -0,0 +1,38 @@
...
+    public T first;
+    public T second;
+
+    public Constraint(T first, T second) {
+        this.first = first;
+        this.second = second;
+    }
+
+    public String toString() {
+        return "Character " + second + " cannot be ordered before " + first;
```
YakimS: Replace "Character" with "Value".  Because T can be a different type.

Copy instead of row 18:
return "Value " + second + " cannot be ordered before " + first;

group/assignment5/Constraint.java
```diff
@@ -0,0 +1,38 @@
+package assignment5;
+
+/**
+ * This class represents a constraint made up of two vertecies of a graph.
+ * First is the vertex from which we can go to the second one.
+ * @param <T>
+ */
+public class Constraint<T extends Comparable<T>> {
+    public T first;
```
YakimS: Change maybe to "former" and "latter"?

group/assignment5/Constraint.java
```diff
@@ -0,0 +1,38 @@
+package assignment5;
+
+/**
+ * This class represents a constraint made up of two vertecies of a graph.
```
YakimS: This class is not about graphs vertices, but representing the relation between 2 Comparable values.

So maybe the class comment should be "This class represents an order-relation constraint between two Comparable objects from the same type"

group/assignment5/Constraint.java
```diff
@@ -0,0 +1,38 @@
...
+        this.first = first;
+        this.second = second;
+    }
+
+    public String toString() {
+        return "Character " + second + " cannot be ordered before " + first;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
YakimS: We should use java conventions in equals(Object o), and hashCode() - use brackets in "if"s, ect.

group/assignment5/Graph.java
```diff
@@ -0,0 +1,218 @@
+package assignment5;
+
+import java.util.Map;
+import java.util.HashMap;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Set;
+import java.util.HashSet;
+import java.util.Collections;
+
```
YakimS: Add static import of "VertexState" to make the code more readable.

copy-paste this:
import static assignment5.VertexState.*;

group/assignment5/Graph.java
```diff
@@ -0,0 +1,218 @@
...
+ * A graph can be described as a set of vertecies. In case graph
+ * is directed each vertex has its indegree and outdegree vertecies.
+ * This class helps you to construct a directed graph and operate on it
+ * easily.
+ */
+public class Graph<T extends Comparable<T>> {
+    private Map<T, Vertex> vertecies;
+
+    public Graph() {
+        vertecies = new HashMap<>();
```
YakimS: Misspelled - use "vertices" instead of "vertecies"

group/assignment5/Graph.java
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * This class represents a directed graph data structure.
+ * A graph can be described as a set of vertecies. In case graph
+ * is directed each vertex has its indegree and outdegree vertecies.
+ * This class helps you to construct a directed graph and operate on it
+ * easily.
+ */
+public class Graph<T extends Comparable<T>> {
+    private Map<T, Vertex> vertecies;
```
YakimS: I'd call it "vertexByValue". Because the map is a lil bit more than just "vertices".

group/assignment5/Graph.java
```diff
@@ -0,0 +1,218 @@
...
+     * @param value is a value of a new vertex
+     */
+    public void addVertex(T value) {
+        if(vertecies.containsKey(value)) {
+            throw new IllegalArgumentException("Vertex " + value + " is already in a graph");
+        }
+        vertecies.put(value, new Vertex(value));
+    }
+
+    public Map<T, Vertex> getVertecies() {
```
YakimS: This function never used. Can be deleted.

group/assignment5/VertexState.java
```diff
@@ -0,0 +1,19 @@
...
+ * Visiting state means that the search from this vertex has already been started
+ * but hasn't been finished yet. This means that we have already visited this vertex
+ * but we are to backtrack to it.
+ * Visited state means that the search started in this vertex has already been finished.
+ *
+ * This enum can also be used to mark vertecies simply as non-visited or
+ * visited if there is no need to track the time when the search from this
+ * vertex has been started and finished which can be done using Visiting state
+ */
+public enum VertexState {
```
YakimS: "DFSState", or "DFSVertexState" is a better name, I think.

group/assignment5/Graph.java
```diff
@@ -0,0 +1,218 @@
...
+
+        public List<Vertex> getIndegrees() {
+            return indegrees;
+        }
+
+        public List<Vertex> getOutdegrees() {
+            return outdegrees;
+        }
+
+        public VertexState getState() {
```
YakimS: "getDFSState()", maybe?

group/assignment5/Graph.java
```diff
@@ -0,0 +1,218 @@
...
+import java.util.Set;
+import java.util.HashSet;
+import java.util.Collections;
+
+
+/**
+ * This class represents a directed graph data structure.
+ * A graph can be described as a set of vertecies. In case graph
+ * is directed each vertex has its indegree and outdegree vertecies.
+ * This class helps you to construct a directed graph and operate on it
```
YakimS: I think this part is redundant:
"This class helps you to construct a directed graph and operate on it easily."

group/assignment5/AlphabetDiscoverer.java
narciuch: Agreed. 
Perhaps also rename 's' to 'currentWord', to make it clearer what the entity is?

group/assignment5/AlphabetDiscoverer.java
```diff
@@ -0,0 +1,86 @@
...
+        int minLength = s1.length() <= s2.length() ? s1.length() : s2.length();
+        for (int i = 0; i < minLength; i++) {
+            if (s1.charAt(i) != s2.charAt(i)) {
+                return i;
+            }
+        }
+        return minLength;
+    }
+
+    public List<Character> getAlphabet(List<String> dictionary) {
```
narciuch: Might be good to add some documentation here, especially since this function is public. A description taken from the specification just to let readers know the expected pre and post conditions of the method. Maybe something like:

```
Given a dictionary (a list of words in lexicographic order) of all words in a possibly fictional language,
this method finds an alphabet (an ordered list of characters) of that language.

This language can contain any character (of the native char data type). 
Upper/lower case characters are treated differently. 
We assume standard lexicographical ordering (order by characters from left to right, if X is a prefix of Y then X is sorted before Y), but where the order of characters is unknown.
```

We could also mention how this function handles the cases of multiple consistent alphabets or no possible alphabets, referencing the getAllPossibleAlphabets function in relation to this?

group/assignment5/Constraint.java
narciuch: Maybe 'from' and 'to' to make it even more explicit?

group/assignment5/Graph.java
```diff
@@ -0,0 +1,218 @@
...
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Set;
+import java.util.HashSet;
+import java.util.Collections;
+
+
+/**
+ * This class represents a directed graph data structure.
+ * A graph can be described as a set of vertecies. In case graph
```
narciuch: I may be wrong, but I think 'vertecies' is misspelled. Isn't it 'vertices'?

group/assignment5/Graph.java
```diff
@@ -0,0 +1,218 @@
...
+     * This class represents a vertex of a graph.
+     * Each vertex can be described by its character-value,
+     * indegree vertecies (list of vertecies that can direct you to the current vertex)
+     * and outdegree vertecies (list of vertecies that can be achieved from the current one)
+     * Vertex state is a special state to track the state of vertex during depth first search
+     * in a graph.
+     */
+    public class Vertex {
+        private T value;
+        private List<Vertex> indegrees;
```
narciuch: Maybe it would be better to use a TreeSet to enforce a sorted collection as well as no duplicate values?

group/assignment5/Constraint.java
YakimS: From-to is about direction, but this class is about order, so I think that former-latter or before-after is better.

group/assignment5/Graph.java
YakimS: Right!

group/assignment5/AlphabetDiscoverer.java
narciuch: Done

group/assignment5/Constraint.java
narciuch: Done

group/assignment5/Constraint.java
narciuch: Done

group/assignment5/Graph.java
narciuch: Done

group/assignment5/Graph.java
narciuch: Done

group/assignment5/Graph.java
narciuch: Done

group/assignment5/Graph.java
narciuch: Done

group/assignment5/AlphabetDiscoverer.java
narciuch: Done by changing in and out vertices into sets

group/assignment5/Graph.java
narciuch: Done using Hashset - the collection in and out degrees don't need to be sorted

group/assignment5/VertexState.java
YakimS: Done

group/assignment5/Graph.java
YakimS: Done

group/assignment5/Graph.java
YakimS: Done

group/assignment5/Constraint.java
YakimS: Done

group/assignment5/Constraint.java
YakimS: Done

group/assignment5/src/VertexState.java
```diff
@@ -1,17 +0,0 @@
...
- * Visiting state means that the search from this vertex has already been started
- * but hasn't been finished yet. This means that we have already visited this vertex
- * but we are to backtrack to it.
- * Visited state means that the search started in this vertex has already been finished.
- *
- * This enum can also be used to mark vertecies simply as non-visited or
- * visited if there is no need to track the time when the search from this
- * vertex has been started and finished which can be done using Visiting state
- */
-public enum VertexState {
```
YakimS: I've moved this class into Graph.java

group/assignment5/src/AlphabetDiscoverer.java
```diff
@@ -17,9 +18,13 @@ private void buildGraphFromDictionary(List<String> dictionary) {
         if(dictionary.isEmpty()) {
             throw new IllegalArgumentException("Dictionary cannot be empty");
         }
-        graph = new Graph<>();
-        addAllCharsInDictionaryToGraph(dictionary);
-        addConstraints(dictionary);
+        if(!dictionary.equals(currentDictionary))
```
YakimS: Created this check so "buildGraphFromDictionary" won't be executed unnecessarily.

group/assignment5/src/Graph.java
```diff
@@ -1,5 +1,10 @@
-import java.util.*;
```
YakimS: Please make sure not to use "*"

group/assignment5/src/Graph.java
```diff
@@ -45,103 +50,101 @@ public void addEdge(T parent, T child) {
...
      * of vertices of the graph
      * @return all topological orders of vertices of a graph
      */
     public List<List<T>> getAllTopologicalOrders() {
-        List<List<T>> alphabets = new ArrayList<>();
-        List<T> alphabet = new ArrayList<>();
-        dfsForAllTopologicalOrders(alphabet, alphabets);
-        vertexByValue.values().forEach(e -> e.setState(VertexState.UNVISITED));
-        return alphabets;
+        List<List<T>> allOrderedVerticesVal = new ArrayList<>();
```
YakimS: I've moved each dfs function next to the caller function so the class will be more readable.

group/assignment5/src/Constraint.java
```diff
@@ -1,36 +1,41 @@
 /**
  * This class represents an order-relation constraint between two Comparable objects from the same type.
- * 'from' is the vertex from which we can go to the second one, 'to'.
- * @param <T>
+ * 'former' is the vertex from which we can go to the second one, 'latter'.
  */
-public class Constraint<T extends Comparable<T>> {
-    public T from;
-    public T to;
+public class Constraint<T> {
```
YakimS: I don't see a good reason T should extend Comparable, so I removed it.

YakimS: If we leave it this way, we need to edit the class comment (remove the word "comparable").

stelaval (approved):
LGTM

PR #30, opened 2018-08-05
Add solution for challenge 1 and 2

yujo-t:


stelaval:
I went through the whole solution. Very nice and clean. Good names for variables ;)

group/assignment6/BiMap/BiMap.java
```diff
@@ -4,66 +4,68 @@
...
 
-    @Override
-    public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
-        BiMap<?, ?> biMap = (BiMap<?, ?>) o;
-        return keyToValue.equals(biMap.keyToValue);
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
```
YakimS: Please use the former definition of equal, so different object with the same content will be considered as equal.

group/assignment6/Graph/Graph.java
```diff
@@ -1,126 +1,116 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
 
 /**
- * This class represents a directed graph data structure.
- * A graph can be described as a set of vertices. In case graph
- * is directed each vertex has its indegree and outdegree vertices.
+ * Represents a graph data structure which is a set of vertices.
```
YakimS: Already mention it on a comment in the commit - mention it is undirected, and removing the implementation detail.

group/assignment6/ParkingLot/ParkingLot.java
```diff
@@ -1,26 +1,30 @@
...
     }
 
     //* Solution 1 */////////////////////////////////////////////////////////
     // Simple to understand, brute-force solution using swaps.  in O(n)
     // Time efficiency is O(n),  Number of car swaps is O(n)
-    public List<CarMove> rearrange(ParkingLot another) throws IllegalArgumentException {
-        if (!isSameParkingLotDiffOrder(another)) {
-            throw new IllegalArgumentException("Not same parking lot in different order");
+    public List<CarMove> rearrange(ParkingLot goalState) throws IllegalArgumentException {
+        if (!isRearrangementOfParkingLot(goalState)) {
```
YakimS, line 69: Better naming, in my opinion, will be: isSameParkingLotInDifferentOrder

group/assignment6/ParkingLot/ParkingLot.java
```diff
@@ -189,8 +260,8 @@ public String toString() {
     public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
-        ParkingLot that = (ParkingLot) o;
-        return Objects.equals(carSpaceBiMap, that.carSpaceBiMap);
+        ParkingLot parkingLot = (ParkingLot) o;
```
YakimS, line 302: "otherParkingLot" instead of only "parkingLot"

group/assignment6/ParkingLot/ParkingLot.java
```diff
@@ -155,6 +159,71 @@ private void initStateMaps(Map<Car, RearrangeState> carStateMap, Map<Space, Spac
...
+            "The goal parking lot state has to be a rearrangement of the initial parking lot state.");
+      }
+
+      List<Move> moves = new ArrayList<Move>();
+      ParkingLot currentState = new ParkingLot(carSpaceBiMap.getKeyToValueMap());
+      moveCars(currentState, goalState, moves);
+      return moves;
+    }
+
+    private static void moveCars(ParkingLot currentState, ParkingLot goalState, List<Move> moves) {
```
YakimS, line 231: Nice! This is a clear and simple algorithm and I think we need to use it for challenge 1 and 2.

group/assignment6/Graph/Graph.java
```diff
@@ -1,126 +1,116 @@
...
+      childVertex = addVertex(child);
     }
 
-
-    public boolean containsVertex(T vertexValue)
-    {
-        return vertexByValue.containsKey(vertexValue);
+    boolean isNewEdgeForChild = childVertex.addAdjacentVertex(parentVertex);
+    boolean isNewEdgeForParent = childVertex.addAdjacentVertex(parentVertex);
+    return  isNewEdgeForChild && isNewEdgeForParent;
```
narciuch: Lines 44 and 45 appear to be calling exactly the same function, so isNewEdgeForChild would always be equal to isNewEdgeFor parent. Why is this value assigned twice and the variables && on 46?

narciuch (approved):
Brilliant work!
Very clear, easy to understand.
Wasn't sure just about that one thing, would appreciate it if you could explain what this does (and maybe add a comment to make it very clear for other code readers too)

group/assignment6/Graph/Graph.java
yujo-t: What do you want to remove?

group/assignment6/BiMap/BiMap.java
yujo-t: I didn't change `equals`. See below.

group/assignment6/Graph/Graph.java
yujo-t: Good catch, I missed that. I think it's supposed to be `parentVertex.addAdjacentVertex(childVertex)`. @YakimS ?

group/assignment6/ParkingLot/ParkingLot.java
yujo-t: Hm, that seems like a mouthful though. Which one do you prefer? `isSameParkingLotInDifferentOrder`, `isRearrangementOfParkingLot` or maybe even `isRearrangementOfInitialState`? @stelaval @narciuch @HelgaShiryaeva

group/assignment6/ParkingLot/ParkingLot.java
yujo-t: I don't think that's necessary actually. I've already put `this.carSpaceBiMap` instead of just `carSpaceBiMap`, plus it's such a trivial function.

group/assignment6/BiMap/BiMap.java
narciuch: Ah yes, it seems git highlighting the differences has made it seem as though you had.

group/assignment6/Graph/Graph.java
narciuch: I agree it's good to mention it's undirected. 
I think Sharon meant remove the  " which is a set of vertices" but I think it's best to leave this in, cause it doesn't go into too much detail but does give the reader a basic understanding of the graph structure and its properties, which is good to know if they are using this class (like the fact that it's a set of vertices so all vertices are unique).

group/assignment6/ParkingLot/ParkingLot.java
YakimS: Better mouthful than unclear: isSameParkingLotInDifferentOrder it is.

group/assignment6/ParkingLot/ParkingLot.java
YakimS: Cool

group/assignment6/ParkingLot/ParkingLot.java
yujo-t: I think it's pretty clear, but let's get more people's opinions.

PR #36, opened 2018-08-05
Final code iterations and documentation

VolhaAnishchanka:


group/assignment6/test/ParkingLotTest.java
```diff
@@ -1,15 +1,11 @@
 package assignment6.test;
 
-import assignment6.ParkingLot.Car;
-import assignment6.ParkingLot.Move;
-import assignment6.ParkingLot.CarMove;
-import assignment6.ParkingLot.ParkingLot;
-import assignment6.ParkingLot.Space;
-
+import assignment6.ParkingLot.*;
```
yujo-t: Don't import *. If we ever add more classes to ParkingLot (not that we will), they will be imported too.

group/assignment6/test/ParkingLotTest.java
```diff
@@ -156,12 +152,10 @@ public void testCanHandleMultipleCircularDependencies() {
...
   private List<List<Move>> getMoveLists(Move[][] moveArrays) {
-    List<List<Move>> moveLists = new ArrayList<List<Move>>();
+    List<List<Move>> moveLists = new ArrayList<>();
     for (Move[] moveArray : moveArrays) {
-      List<Move> moveList = new ArrayList<Move>();
-      for (Move move : moveArray) {
-        moveList.add(move);
-      }
+      List<Move> moveList = new ArrayList<>();
+      Collections.addAll(moveList, moveArray);
```
yujo-t, line 162: Nice.

group/assignment6/test/ParkingLotTest.java
```diff
@@ -173,9 +167,9 @@ public void testEmptyParkingLot() {
     ParkingLot goalState = new ParkingLot(goalMap);
 
     List<Move> moves = initialState.getCarMoves(goalState);
-    List<Move> expectedMoves = new ArrayList<Move>();
+    List<Move> expectedMoves = new ArrayList<>();
 
-    assertTrue(expectedMoves.equals(moves));
+    assertEquals(expectedMoves, moves);
```
yujo-t, line 176: Does this work for Lists?

group/assignment6/test/ParkingLotTest.java
```diff
@@ -194,13 +188,6 @@ public void allRearrangementsTest() {
...
-    System.out.println(allMoves.size());
-    for (List<CarMove> carMoves : allMoves) {
-      if(carMoves.size() == 3) {
-        System.out.println("YEEEAH, sequence of moves with the least amount if moves is also here !");
-      }
-    }
-    //GIRLS, IT WOOOOOOOOOOOOOOOOOOOOOOOOORKS !!
     for (List<CarMove> carMoves : allMoves) {
       initialState = new ParkingLot(initialMap);
       initialState.makeMoves(carMoves);
```
yujo-t, line 197: Could you check whether each move sequence is actually what we expect it to be? See the other tests for examples. You can reduce the parking lot size for simplicity.

group/assignment6/ParkingLot/Move.java
```diff
@@ -3,7 +3,7 @@
 import java.util.Objects;
 
 /**
- * Represents a move of a car.
+ * This class represents a move of a car.
```
yujo-t: From looking at the javadocs for builtin classes, it looks like the convention is to start with a verb, therefore `Represents a...` would be better.

group/assignment6/ParkingLot/ParkingLot.java
```diff
@@ -1,18 +1,22 @@
...
+
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Objects;
 
-import assignment6.BiMap.BiMap;
-
+/**
+ * This class represents a Parking Lot. Provides
```
yujo-t, line 11: Change to `Represents a parking lot that can have spaces in which cars can be parked.`. The efficiency bit is not needed.

group/assignment6/ParkingLot/ParkingLot.java
```diff
@@ -119,19 +122,25 @@ private static boolean carIsInWrongSpace(Car car, ParkingLot currentState, Parki
 
   // ================================== Solution to challenge 4 ==================================
 
+  /**
+   * Method for getting all possible unique sequences of moves of cars in
```
yujo-t: `Gets all possible unique sequences of moves of cars in a parking lot. Each sequence will not contain the same parking lot configuration more than once.`

group/assignment6/ParkingLot/ParkingLot.java
```diff
@@ -119,19 +122,25 @@ private static boolean carIsInWrongSpace(Car car, ParkingLot currentState, Parki
 
   // ================================== Solution to challenge 4 ==================================
 
+  /**
+   * Method for getting all possible unique sequences of moves of cars in
+   * a Parking Lot. Each sequence will not contain same Parking Lot configuration
+   * during rearrangement.
+   * @param another is a ParkingLot end configuration.
```
yujo-t: Could you rename `another` to `goalState`?

group/assignment6/ParkingLot/ParkingLot.java
```diff
@@ -119,19 +122,25 @@ private static boolean carIsInWrongSpace(Car car, ParkingLot currentState, Parki
...
-    allRearrangementsRecursive(allMoves, currentMoves, previousConfigurations, another, null);
+    allRearrangementsRecursive(allMoves, currentMoves, previousConfigurations, another);
     carSpaceBiMap = new BiMap<>(currentParkingLot);
     return allMoves;
   }
 
   private void allRearrangementsRecursive(List<List<CarMove>> allMoves, List<CarMove> currentMoves,
-      List<ParkingLot> previousConfigurations, ParkingLot desired,
-      Car previous) {
+                                          List<ParkingLot> previousConfigurations, ParkingLot desired) {
```
yujo-t: Multi-line parameters should be indented 4 spaces.

group/assignment6/ParkingLot/ParkingLot.java
```diff
@@ -188,7 +197,7 @@ private CarMove makeMove(Car car, Space from, Space to) {
     return null;
   }
 
-  public void moveCarToEmptySpace(Car car) {
+  private void moveCarToEmptySpace(Car car) {
```
yujo-t, line 202: If you have time, could you have a look at my implementation of `moveCarToEmptySpace` and see if they can be merged?

group/assignment6/ParkingLot/ParkingLotWithReservedSpaces.java
```diff
@@ -1,180 +1,180 @@
...
-
-        return Objects.hash(carSpaceBiMap);
-    }
-}
+package assignment6.ParkingLot;
+
+import assignment6.BiMap.BiMap;
+import assignment6.Graph.Graph;
+
+import java.util.*;
```
yujo-t: Don't import *

group/assignment6/ParkingLot/ParkingLotWithReservedSpaces.java
```diff
@@ -1,180 +1,180 @@
...
+        carSpaceBiMap.put(to, car);
+        if (!car.equals(Car.noCar)) {
+            return new CarMove(car,from,to);
+        }
+        return null;
+    }
+
+
+    public static boolean areMovesCreateEndResult(ParkingLotWithReservedSpaces start,
+                                                  ParkingLotWithReservedSpaces end,
```
yujo-t, line 147: Indent 4 spaces

yujo-t (changes requested):

group/assignment6/test/ParkingLotTest.java
VolhaAnishchanka: Yeah, it works )

group/assignment6/ParkingLot/ParkingLotWithReservedSpaces.java
VolhaAnishchanka: I cannot understand indenting by four spaces. Could you give an example ?

group/assignment6/test/ParkingLotTest.java
yujo-t: Nice :D 🎉

group/assignment6/ParkingLot/ParkingLotWithReservedSpaces.java
yujo-t: So the second line should start 4 spaces from where `public` starts.
```
public static...
    ParkingLot....
```

group/assignment6/test/ParkingLotTest.java
VolhaAnishchanka: But I have 2921 unique sequences there, it is impossible, I am sure this assertion is enough in this case

yujo-t: What if you reduce the parking lot to just 2 cars and 3 spaces?

group/assignment6/ParkingLot/ParkingLot.java
VolhaAnishchanka: Why they cannot be merged ? I just changed public to private because we didn't use it outside the class )

yujo-t: Changing it to private is fine. It's just that there are two `moveCarToEmptySpace` methods right now in `ParkingLot`. One you're using and one I'm using.

yujo-t (approved):

